PR #7, opened 2019-02-18
add editing feature to users

nivemaham:
- allow editing of `projectId`, `startDate` and `EndDate`.
- uses `Date` instead of custom model in Datepicker


blootsvoets:
LGTM. Good that the backend code was already there. Small remark about the backend: the user is data not completely replaced, so probably a POST request would be more to the point than a PUT request. That's not blocking though.

PR #15, opened 2019-05-15
Add support for pagination, sorting and filtering

### LINKED ISSUE #12: Enable Pagination
Enable Pagination for the list of users in the UI.

### LINKED ISSUE #14: Add ability to sort, filter by Project ID
This is requested by the clinical sites project admins for easily checking the authorised participants.

yatharthranjan:
Adds support for Client side pagination, sorting and filtering.
- 

NOTE: For better performance this should be implemented in the future in backend. #17 

- Closes #14 
- Closes #12 
- Adds pagination, sorting and filtering using mat-table, mat-paginator and mat-form-field
- One form for filtering for any column in the table
- Adds material theme

Here is a gif of how it looks -

radar-auth-example

### DIFF
--- a/authorizer-app/src/app/components/rest-source-authorization/rest-source-user-list.component.html
+++ b/authorizer-app/src/app/components/rest-source-authorization/rest-source-user-list.component.html
@@ -1,40 +1,74 @@
 <div>
   <ngb-alert *ngIf="errorMessage" type="danger" (close)="errorMessage = null">{{ errorMessage}}</ngb-alert>
   <rest-source-user-registration-form></rest-source-user-registration-form>
-  <div class="container table-responsive"  *ngIf="restSourceUsers">
-    <table class="table table-hover table-striped ">
-      <thead>
-      <tr>
-        <th><span>Id</span></th>
-        <th><span>Project Id</span></th>
-        <th><span>User Id</span></th>
-        <th><span>Source Id</span></th>
-        <th><span>Start Date</span></th>
-        <th><span>End Date</span></th>
-        <th><span>External User Id</span></th>
-        <th><span>Authorized</span></th>
-        <th></th>
-        <th></th>
-      </tr>
-      </thead>
-      <tbody>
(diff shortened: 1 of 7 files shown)


authorizer-app/src/app/components/rest-source-authorization/rest-source-user-list.component.html
```diff
@@ -1,40 +1,74 @@
...
+            <button mat-raised-button	color="warn" (click)="removeDevice(row)">Delete</button>
+        </mat-cell>
+      </ng-container>
+
+      <mat-header-row *matHeaderRowDef="displayedColumns; sticky: true"></mat-header-row>
+      <mat-row *matRowDef="let row; columns: displayedColumns;">
+      </mat-row>
+    </mat-table>
+
+  <mat-paginator [pageSize]="5" [pageSizeOptions]="[5, 10, 20, 50]" showFirstLastButtons></mat-paginator>
```
blootsvoets: Perhaps set the default to 20 per page?

nivemaham:
Nice work. Although I prefer to have the pagination supported from back-end as well. Pagination with Spring is fairly straight forward. Could you take that up @yatharthranjan

yatharthranjan:
@blootsvoets No updated needed for `package.json`. 

@nivemaham I looked into it when I started implementing this and I think it will require some work and probably for another PR. As it will involve adding all the 3 (pagination, sorting and filtering) from the backend and then also creating custom datasources in the frontend as `MatTableDataSource` only supports client side pagination, sorting and filtering. For the front end part, see this post for an example -- https://blog.angular-university.io/angular-material-data-table/ 
I will create an issue for this and maybe try to do it another PR but this should be sufficient for now with the number of rows we have

yatharthranjan:
Hi Guys can we proceed with this please ? I have created issue for server side support #17

yatharthranjan:
Yes that is because I ran `npm update` and caused updated versions in `package-lock.json` for dependencies with Caret `^` and Tilde `~` in `package.json` ?

Issue #13, opened 2019-05-13
Add integration with the Management Portal


yatharthranjan:
Integrate the Authoriser with the management portal in the first instance to - 

- [ ] Validate the project Ids and subject Ids (If they exist in the Management Portal)
- [ ] Later we can use the same Mp client (and exposed end point of the backend) to create a drop down in the UI to select the project Id and subject Id.

In the second instance, we can 

- [ ] Add auth from the management portal so the UI only shows the projects and subjects of which a user has authorization for.
- [ ] Integrate source registration to management portal and link back to the authorizer.

PR #25, opened 2019-10-03
Add validation

yatharthranjan:
**FrontEnd**:
- Adds a simple regex-based validation (based on ids generated by MP) to the frontend and some sanity checks to avoid silly mistakes.
- Add builder args to docker file so can easily change in a development environment
- Add VALIDATE to the environment and use it to make user update requests with backend validation. Parses and displays errors correctly.

**BackEnd**:
- Adds a simple MP client and Validator to validate the requests.
- Specify the type of validator in the properties (currently only `managementportal` available).
- Adds a `validate=true/false` request parameter to the update users endpoint.

Partially Solves #13


authorizer-app/src/app/services/rest-source-user.service.ts
```diff
@@ -10,27 +10,29 @@ import {environment} from "../../environments/environment";
...
+  }
 
   getAllUsers(): Observable<RestSourceUser[]> {
     return this.http.get<RestSourceUser[]>(this.serviceUrl);
   }
 
   updateUser(sourceUser: RestSourceUser): Observable<any> {
-
-    return this.http.post(this.serviceUrl + '/' + sourceUser.id, sourceUser);
+    const params = new HttpParams().set("validate", String(environment.VALIDATE));
```
nivemaham: why not validate for all environments?

authorizer-app/src/index.html
```diff
@@ -3,7 +3,7 @@
 <head>
   <meta charset="utf-8">
   <title>AuthorizerApp</title>
-  <base href="/rest-sources/authorizer/">
+  <base href="/">
```
nivemaham: I think you have to revert this for docker build

build.gradle
```diff
@@ -35,6 +35,7 @@ ext {
 	description = 'RADAR Rest Source Authorizer handles authorization for data access from third party APIs for wearable devices or other connected sources.'
 
 	okhttp3Version = '3.13.1'
+	managementPortalVersion = '0.5.7'
```
nivemaham: You can go for `MP 0.5.8` since it is already available.

src/main/java/org/radarbase/authorizer/service/managementportal/CachedManagementPortalClient.java
```diff
@@ -0,0 +1,218 @@
...
+  public Project getProject(String projectId) throws IOException, TokenException {
+    if (isUpdateRequired()) {
+      update();
+      return this.projects.stream()
+          .filter(project1 -> project1.getProjectId().equals(projectId))
+          .findFirst()
+          .orElse(null);
+    }
+    Optional<Project> project = this.projects.stream()
+        .filter(project1 -> project1.getProjectId().equals(projectId))
```
nivemaham: why not find?

src/main/java/org/radarbase/authorizer/service/managementportal/CachedManagementPortalClient.java
```diff
@@ -0,0 +1,218 @@
...
+  @Override
+  public Project getProject(String projectId) throws IOException, TokenException {
+    if (isUpdateRequired()) {
+      update();
+      return this.projects.stream()
+          .filter(project1 -> project1.getProjectId().equals(projectId))
+          .findFirst()
+          .orElse(null);
+    }
+    Optional<Project> project = this.projects.stream()
```
nivemaham: Is it this redundant? :thinking:

src/main/java/org/radarbase/authorizer/service/managementportal/CachedManagementPortalClient.java
```diff
@@ -0,0 +1,218 @@
...
+    this.oAuth2Client = new OAuth2Client.Builder()
+        .credentials(properties.getOauthClientId(), properties.getOauthClientSecret())
+        .endpoint(new URL(properties.getBaseUrl()), "/oauth/token")
+        .httpClient(httpClient)
+        .build();
+  }
+
+  private ManagementPortalProperties loadManagementPortalProperties(@NotNull String path)
+      throws ConfigurationException {
+    ManagementPortalProperties properties = ConfigHelper
```
nivemaham: little confused here with why using YAML models and using property style implementation. I think you can also annotate @NotNull to the class definition. Then it will automatically validate when loading. so no need to do the following check.

src/main/java/org/radarbase/authorizer/webapp/resource/RestSourceUserResource.java
```diff
@@ -35,79 +37,90 @@
...
-        logger.debug("Requesting to delete rest source user");
-        this.restSourceUserService.revokeTokenAndDeleteUser(Long.valueOf(id));
-        return ResponseEntity
-                .ok().header("user-removed", id).build();
+  @PostMapping("/users/{id}")
+  public ResponseEntity updateDeviceUser(@Valid @PathVariable String id,
+      @RequestBody RestSourceUserPropertiesDTO restSourceUser,
+      @RequestParam(value = "validate", defaultValue = "false") Boolean isValidate) {
+    logger.debug("Requesting to update rest source user");
+    if (validator != null && isValidate && !validator.validate(restSourceUser)) {
```
nivemaham: Minor detail. I would switch the `isValidate` first then check if `validator != null`

src/main/java/org/radarbase/authorizer/webapp/resource/RestSourceUserResource.java
```diff
@@ -35,79 +37,90 @@
...
+  }
 
-    @DeleteMapping("/users/{id}")
-    public ResponseEntity<Void> deleteDeviceUser(@Valid @PathVariable String id) {
-        logger.debug("Requesting to delete rest source user");
-        this.restSourceUserService.revokeTokenAndDeleteUser(Long.valueOf(id));
-        return ResponseEntity
-                .ok().header("user-removed", id).build();
+  @PostMapping("/users/{id}")
+  public ResponseEntity updateDeviceUser(@Valid @PathVariable String id,
```
nivemaham: So why do we validate only on update, but not on other operations?

authorizer-app/src/index.html
yatharthranjan: I added base href as an argument to the docker build so the path should be added then i think

src/main/java/org/radarbase/authorizer/service/managementportal/CachedManagementPortalClient.java
yatharthranjan: yes was planning to use streams at first but decided against it. Now using simple hashsets

src/main/java/org/radarbase/authorizer/webapp/resource/RestSourceUserResource.java
yatharthranjan: I guess because there is not much to validate in the other ones? If there is, i will be happy to add it to them

authorizer-app/src/app/services/rest-source-user.service.ts
yatharthranjan: I think this will get the value for `VALIDATE` based on the current environment.

src/main/java/org/radarbase/authorizer/service/managementportal/CachedManagementPortalClient.java
yatharthranjan: Do you mean that i should use application.properties for management portal properties instead of a separate file?
Ok good point will add the NotNull checks.

src/main/java/org/radarbase/authorizer/service/managementportal/CachedManagementPortalClient.java
yatharthranjan: The stream doesn't have a find(predicate) function. Am i looking in the worng place?

src/main/java/org/radarbase/authorizer/service/managementportal/CachedManagementPortalClient.java
yatharthranjan: We can decrease the fetch interval for the cache to something like 10-30 seconds but i dont want it querying all the subjects too much since the list of subjects can be quite long.

src/main/java/org/radarbase/authorizer/service/managementportal/CachedManagementPortalClient.java
nivemaham: Hm. Not exactly. But what you said is even better. read these properties from Application.properties file. So no need to add another config file. and spring application file works with ENV variables for docker, which makes the config easier

src/main/java/org/radarbase/authorizer/service/managementportal/CachedManagementPortalClient.java
nivemaham:
```
@Override
  public Project getProject(String projectId) throws IOException, TokenException {
    if (isUpdateRequired()) {
      update();
    }
    var project = this.projects.stream()
          .filter(project1 -> project1.getProjectId().equals(projectId))
          .findFirst()
          .orElse(null);

  return project.orElse(queryProject(projectId));
}
```
I meant something like this. I understand the reason behind the cache and fetch when not available.

authorizer-app/src/app/services/rest-source-user.service.ts
nivemaham: Yes. But i am curious why do we optionally validate instead of always?

yatharthranjan: For example, we might not want to validate in a development environment. Just to make it more configurable

authorizer-app/src/index.html
yatharthranjan: I changed it back but can still be controlled by docker ARGS

build.gradle
yatharthranjan: Done

src/main/java/org/radarbase/authorizer/service/managementportal/CachedManagementPortalClient.java
yatharthranjan: I made it more readable. Hope it makes more sense now

src/main/java/org/radarbase/authorizer/service/managementportal/CachedManagementPortalClient.java
yatharthranjan: Done

src/main/java/org/radarbase/authorizer/webapp/resource/RestSourceUserResource.java
yatharthranjan: Done

.gitignore
```diff
@@ -25,4 +26,5 @@
 /nbdist/
 /.nb-gradle/
 
-/src/main/docker/etc/rest_source_clients_configs.yml
\ No newline at end of file
+/src/main/docker/etc/rest_source_clients_configs.yml
+/src/main/docker/etc/management_portal_config.yml
```
nivemaham: You can remove this line

PR #34, opened 2020-01-14
Fixes for URL paths

yatharthranjan:
- also verifies the connection and token privileges on start of the application (helps report and diagnose problems before start using it)
-  makes token path configurable

This is needed because of the `URL` constructor `URL(context, spec)`. If the spec has a `/` in the begining then it will replace the path in the context. Hence when creating `new URL("http://managementportal-app:8080/managementportal", "/oauth/token")`, the `/managementportal` path was being discarded. see [official docs](https://docs.oracle.com/javase/7/docs/api/java/net/URL.html#URL(java.net.URL,%20java.lang.String)) for more info

Migration Steps involving validation
-

### If don't need validation
Add the `REST_SOURCE_AUTHORIZER_VALIDATOR` env var to your docker-compose service to disable validation-
```yaml
  radar-rest-sources-backend:
    image: radarbase/radar-rest-source-auth-backend:1.2.1
...
    environment:
...
      - REST_SOURCE_AUTHORIZER_VALIDATOR=""
    volumes:
      - ./etc/rest-source-authorizer/:/app-includes/
...

```
**Note: This will only disable backend validation. The frontend validation(based on Regex) will still exist.**

#### First Create a new oAuth client in Management Portal
To add new OAuth clients, you can add at runtime through the UI on Management Portal, or you can add them to the OAuth clients file referenced by the MANAGEMENTPORTAL_OAUTH_CLIENTS_FILE configuration option.

#### Then add the following to your rest authoriser service
Add the following env vars to your docker-compose service-
```yaml
  radar-rest-sources-backend:
    image: radarbase/radar-rest-source-auth-backend:1.2.1
...
    environment:
...
      - REST_SOURCE_AUTHORIZER_VALIDATOR=managementportal
      - REST_SOURCE_AUTHORIZER_MANAGEMENT_PORTAL_BASE_URL=http://managementportal-app:8080/managementportal/
      - REST_SOURCE_AUTHORIZER_MANAGEMENT_PORTAL_OAUTH_CLIENT_ID=radar_rest_sources_auth
      - REST_SOURCE_AUTHORIZER_MANAGEMENT_PORTAL_OAUTH_CLIENT_SECRET=secret
    volumes:
      - ./etc/rest-source-authorizer/:/app-includes/
...
```

**Note**: Make sure to configure the client id and client secret as created in the Management portal

### DIFF
--- a/src/main/java/org/radarbase/authorizer/service/managementportal/CachedManagementPortalClient.java
+++ b/src/main/java/org/radarbase/authorizer/service/managementportal/CachedManagementPortalClient.java
@@ -21,11 +21,13 @@ import org.radarbase.authorizer.config.RestSourceAuthorizerProperties;
 import org.radarbase.authorizer.service.dto.managementportal.Project;
 import org.radarbase.authorizer.service.dto.managementportal.Subject;
 import org.radarcns.exception.TokenException;
+import org.radarcns.oauth.OAuth2AccessTokenDetails;
 import org.radarcns.oauth.OAuth2Client;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.http.HttpHeaders;
 import org.springframework.stereotype.Service;
 
 @Service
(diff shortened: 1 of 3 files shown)


src/main/java/org/radarbase/authorizer/service/managementportal/CachedManagementPortalClient.java
```diff
@@ -146,7 +169,7 @@ private Subject querySubject(String subjectId) throws IOException, TokenExceptio
 
   private Project queryProject(String projectId) throws IOException, TokenException {
     Project project = queryEntity(
-        properties.getBaseUrl() + "/api" + properties.getProjectsPath() + "/" + projectId,
+        properties.getBaseUrl() + properties.getProjectsPath() + "/" + projectId,
```
nivemaham: why do we need to configure project paths etc ? can't we just configure managementportal baseUrl and you add the string here?

src/main/java/org/radarbase/authorizer/service/managementportal/CachedManagementPortalClient.java
```diff
@@ -173,14 +197,14 @@ private Project queryProject(String projectId) throws IOException, TokenExceptio
 
   private Set<Subject> queryAllSubjects() throws IOException, TokenException {
     // get subjects from MP
-    return queryEntity(properties.getBaseUrl() + "/api" + properties.getSubjectsPath(),
+    return queryEntity(properties.getBaseUrl() + properties.getSubjectsPath(),
```
nivemaham: Why don't we need `api` here? :thinking:

src/main/resources/application.yml
```diff
@@ -25,8 +25,9 @@ rest-source-authorizer:
 
   validator: managementportal
   management-portal:
-    base-url: "http://localhost:8081"
-    projects-path: "/projects"
-    subjects-path: "/subjects"
+    base-url: "http://localhost:8081/"
+    projects-path: "api/projects"
```
nivemaham: ok. Now I get it. I think this doesn't necessarily need to be a config. 
Paths in ManagementPortal won't change that easily. This is fine for now, since this doesn't need to be configured by user. Otherwise, this will make the configuration process rather difficult.

PR #40, opened 2020-01-22
Initial implementation for RESET of users

### LINKED ISSUE #33: Add RESET capability for re-pulling data
we need a RESET button which will allow for the connector to pull all the historic data for the user again.
As specified in https://github.com/RADAR-base/RADAR-REST-Connector/pull/50 which creates a unique id using the existing user id and a new version (`new id = user id + # + version` if version is not null), we need to add a `version` property to the user and increment it each time it is RESET. For the first instance, this should be `null`, so as to make it backward compatible with the ids in the current connector offsets.

yatharthranjan:
- Adds a version property to user details that is the current time instant as a ISO format string.
- This will enable re-pull of user's data.

Working demo is [here](https://radar-backend.co.uk/rest-sources/authorizer/)
Closes #33

### DIFF
--- a/authorizer-app/package-lock.json
+++ b/authorizer-app/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "authorizer-app",
-  "version": "1.0.0",
+  "version": "1.2.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -11,7 +11,7 @@
       "dev": true,
       "requires": {
         "@angular-devkit/core": "0.8.4",
-        "rxjs": "6.2.2"
+        "rxjs": "~6.2.0"
       }
     },
     "@angular-devkit/build-angular": {
@@ -42,7 +42,7 @@
         "loader-utils": "1.1.0",
         "mini-css-extract-plugin": "0.4.4",
         "minimatch": "3.0.4",
-        "node-sass": "4.11.0",
+        "node-sass": "^4.9.3",
         "opn": "5.4.0",
         "parse5": "4.0.0",
         "portfinder": "1.0.19",
@@ -160,10 +160,10 @@
       "integrity": "sha512-oqMHezbIZMApud9JZDupWaxJeczTA17hLFGJ1qyAaPBRADtjnuguygXLcBLzYYAhzHKstrHwPJ4R1jj3oG28Ow==",
       "dev": true,
       "requires": {
-        "ajv": "6.4.0",
-        "chokidar": "2.0.4",
-        "rxjs": "6.2.2",
-        "source-map": "0.5.7"
+        "ajv": "~6.4.0",
+        "chokidar": "^2.0.3",
+        "rxjs": "~6.2.0",
+        "source-map": "^0.5.6"
       }
     },
     "@angular-devkit/schematics": {
@@ -173,7 +173,7 @@
       "dev": true,
       "requires": {
         "@angular-devkit/core": "0.8.4",
-        "rxjs": "6.2.2"
+        "rxjs": "~6.2.0"
       }
     },
     "@angular/animations": {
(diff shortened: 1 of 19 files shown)


afolarin:
Looks fine to me. 

If applicable to individual streams/topics -- this could be handy, but not essential (at least I don't think we're hitting the API limits at the moment). Certainly, we'd like to re-pull all data for individual participants with detected gaps in their data, or those exiting the study, as a way of trying to backfill any holes in the user data where there was a sync delay on the Fitbit-side (this is the working hypothesis for gaps). 

On a related matter -- is there any reasonable sense to adding a backoff if we detect a given individual's data is gappy in the initial collection phase?

yatharthranjan:
> If applicable to individual streams/topics -- this could be handy, but not essential (at least I don't think we're hitting the API limits at the moment). Certainly, we'd like to re-pull all data for individual participants with detected gaps in their data, or those exiting the study, as a way of trying to backfill any holes in the user data where there was a sync delay on the Fitbit-side (this is the working hypothesis for gaps).

I looked into doing this per topic but it is not as trivial as it would seem. Firstly, the source-authoriser has no data about the topics and data streams so adding them here would take some thinking and further implementation (might not be worth it). Secondly, the connector also will have to be modified to filter this based on the topic and again would require additional effort to design and implement. I guess for now this should be sufficient. We can discuss in the next dev call for potential solutions

> On a related matter -- is there any reasonable sense to adding a backoff if we detect a given individual's data is gappy in the initial collection phase?

I think there already is a back off but it is hard to set this interval since it will vary with every user and how they use the device. Also will vary per data stream. But again we should discuss this in the dev call next week. Perhaps @blootsvoets can also comment his thoughts on this.

nivemaham:
Can't we append some incremented value to the version? so that we have some idea how often this has been reset? Ideally it would be nice to have auditing implemented for this, so we know whats happening. 
@yatharthranjan I also think, it is good to add a confirmation dialog to let user know and confirm that they want to reset.
In another project, we implemented a pop-up which would confirm the operation and you can also optionally choose the start and end date for the reset. in this case we have more control over what data is being pulled and we can maybe avoid redundant pulling.
This would address part of the requirements mentioned by @afolarin

yatharthranjan:
Hi nivethika,

> Can't we append some incremented value to the version? so that we have some idea how often this has been reset? Ideally it would be nice to have auditing implemented for this, so we know whats happening.

Do you mean like {number}-{Date} and the number is also incremented every time we reset? Maybe we can include another field like `timesReset` for brevity instead of adding it to the version? 
By auditing, do you mean auto-generated createDate and updateDate properties?

> @yatharthranjan I also think, it is good to add a confirmation dialog to let user know and confirm that they want to reset.

Ok i will add a pop to both RESET and DELETE.

> In another project, we implemented a pop-up which would confirm the operation and you can also optionally choose the start and end date for the reset. in this case we have more control over what data is being pulled and we can maybe avoid redundant pulling.
This would address part of the requirements mentioned by @afolarin

ok sounds good. But this could be

afolarin:
should we have a cancel on the `Edit` screen seems to only have a `Submit`

yatharthranjan:
Hi @nivemaham and @afolarin ,

I have made most of the requested changes including-

1. adding a timesReset property to signify the number of times a user has been reset
2. Pop up confirmation for the RESET and DELETE Actions
3. The RESET popup also contains functionality to update the start and end dates for data collection.
4. Add a CANCEL button to the EDIT users page.

The only thing I have not yet added is the auditing as i was not exactly sure of the specifics. But it is another feature and can be added in a separate PR.

The live demo for the above changes is available [here](https://radar-backend.co.uk/rest-sources/authorizer/users). Please take a look and provide feedback. Thanks.

Issue #32, opened 2019-12-09
Add authorization with Management portal using `authorization-code`


nivemaham:
We need a systematic way to authenticate users for rest-auth.
We have to add support to login with Management Portal auth using `authorization-code` flow. 
@mpgxvii would you be able to take a look at the front-end side?
@yatharthranjan @blootsvoets Have we done anything on this topic? I guess not yet. Am I right?

yatharthranjan:
@nivemaham no we have not made any start on this yet but it is listed as one of the points in issue #13

yatharthranjan:
We can use the [radar-spring](https://github.com/RADAR-base/radar-spring) library to provide authorization in the backend.

blootsvoets:
This was fixed with #55.

PR #41, opened 2020-01-28
Add authorization with Management portal 

mpgxvii:
- Adds authorization with Management Portal using `authorization-code`, solves #32


authorizer-app/src/app/services/auth.service.ts
```diff
@@ -0,0 +1,87 @@
...
+import { HttpClient, HttpHeaders, HttpParams } from '@angular/common/http';
+
+import { Injectable } from '@angular/core';
+import { environment } from 'src/environments/environment';
+import { map } from 'rxjs/operators';
+import { of } from 'rxjs';
+import { storageItems } from '../enums/storage';
+
+@Injectable()
+export class AuthService {
```
yatharthranjan: Can we create an interface called `AuthService` and then this will implement that and perhaps be called `ManagementPortalAuthService` or if more generic then `AuthorizationCodeAuthService`. This will make it easy to extend in the future and since it is already injected we can provide whichever AuthService we like.

mpgxvii: Okay sure 👍

authorizer-app/src/app/services/auth.service.ts
```diff
@@ -1,87 +1,52 @@
...
+  clearAuth() {
+    this.clearAccessToken();
+    this.clearUser();
   }
 
-  logout() {
-    return of(true);
+  authenticate(loginParams): Observable<any> {
+    throw new Error('AuthService method not implemented');
   }
```
yatharthranjan: Why not use an abstract class? And make this function an abstract function.

README.md
```diff
@@ -29,4 +29,48 @@ To build and run this application from Docker:
 docker build -t radarbase/radar-rest-source-auth-backend:1.0.1 .
 
 docker run -p 8080:8080 radarbase/radar-rest-source-auth-backend:latest
-```
\ No newline at end of file
+```
+
+## Validation
```
nivemaham, line 34: Could you please add documentation for setting the env variable "AUTHORIZATION_CODE"

authorizer-app/src/environments/environment.prod.ts
```diff
@@ -1,5 +1,13 @@
 export const environment = {
   production: true,
   BACKEND_BASE_URL: '/rest-sources/backend',
-  VALIDATE: true
+  VALIDATE: true,
+  AUTH: {
```
nivemaham, line 5: would this work on dockerized env as well? 
Could you please add this documentation to README as well

PR #42, opened 2020-01-28
Adds filter on the front end for users

mpgxvii:
- Adds filter on the front end to hide rest source user data from users who do not have the correct permissions from management portal (uses management portal assigned projects)
- I guess this would be better implemented on the back end?

Partially solves #13

### DIFF
--- a/authorizer-app/package.json
+++ b/authorizer-app/package.json
@@ -34,7 +34,7 @@
     "bootstrap": "^4.1.3",
     "core-js": "^2.5.4",
     "font-awesome": "^4.7.0",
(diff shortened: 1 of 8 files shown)


yatharthranjan:
hi @mpgxvii , I think yes this needs to be done from the backend based on the token provided by the frontend. The frontend should only be responsible for getting a valid token using auth code flow.

mpgxvii:
@yatharthranjan Also, I was wondering maybe it would be good to implement this similar to Management Portal front end? Where projects (where the user has access to) are shown upon login?

PR #53, opened 2020-06-04
Authenticate endpoints

nivemaham:
- Adds a simple Authentication Filter to authenticate endpoints and validate token from Management Portal
- Adds a simple health check resource and exclude from Filters

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,8 @@ To install functional RADAR-base Rest-Sources Authorizer application with minima
 ```bash
 docker-compose up -d --build
 ```
-
+You can find the Authorizer app running on `http://localhost:8080/rest-sources/authorizer/`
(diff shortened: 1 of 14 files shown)


yatharthranjan:
Why not filter the subjects (in the controller method) based on the permissions on the token? So only those subjects for which the user has access to will be displayed in the frontend then?

nivemaham:
This is simply an attempt to complete the authentication with MP. The frontend does not query subjects or projects based on user permission. I agree that it would be the way to go. But it would need some work on both ends. This is an attempt to get the authorizer work with MP authentication and use a token from MP instead of Basic authentication on nginx.

nivemaham:
@mpgxvii Sounds good to me.

PR #54, opened 2020-06-04
Separate MP/auth-code auth and simple/no auth

mpgxvii:
- Add simple auth service to allow no auth (similar to how it was done previously, and just using nginx auth from nginx side)
- Separate simple login and mp login components since the two flows are quite different
- Add factory to determine which auth type to use from environment variables


yatharthranjan:
Shouldn't this be done from the backend side? Since the backend will restrict requests without a valid token?

PR #65, opened 2020-08-31
Migrate to Grizzly and kotlin

nivemaham:
Migrate to Grizzly and Kotlin

TODO

- [x] Migrate dockerfile
- [x] Code clean-up 

I would like to work on the tests on a separate PR (hopefully a bit later?)


blootsvoets:
Regarding code formatting: Kotlin code style is 4 spaces https://kotlinlang.org/docs/reference/coding-conventions.html#formatting

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/HeathCheckResource.kt
```diff
@@ -0,0 +1,23 @@
...
+@Path("health")
+@Resource
+@Singleton
+class HealthCheckResource(
+    @Context private var userRepository: RestSourceUserRepository
+) {
+  @GET
+  fun check(): Response {
+    userRepository.query(Page(0, 1))
+    return Response.ok("Initialized dao and made a sample query...").build()
```
blootsvoets, line 21: Perhaps return a simple data class here?
```
val status = try {
    userRepository.query(Page(0, 1))
    HealthStatus("UP")
} catch (ex: Throwable) {
    HealthStatus("DOWN")
}
return Response.ok(status).build()

data class HealthStatus(val status: String)

File: authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/ProjectResource.kt [hunk 2]
```diff
@@ -0,0 +1,45 @@
...
+    @Context private val auth: Auth
+) {
+
+  @GET
+  @NeedsPermission(Permission.Entity.PROJECT, Permission.Operation.READ)
+  fun projects() = ProjectList(projectService.userProjects(auth))
+
+  @GET
+  @Path("{projectId}/users")
+  @NeedsPermission(Permission.Entity.PROJECT, Permission.Operation.READ, "projectId")
```
blootsvoets, line 34: I probably did this wrong on the original project, but this should probably request for the `SUBJECT` `READ` permission in `projectId`.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RestSourceUserResource.kt
```diff
@@ -0,0 +1,154 @@
...
+      @DefaultValue("1") @QueryParam("page") pageNumber: Int): RestSourceUsers {
+
+    if (projectId != null) {
+      auth.checkPermissionOnProject(Permission.PROJECT_READ, projectId)
+    }
+
+    val queryPage = Page(pageNumber = pageNumber, pageSize = pageSize)
+    val (records, page) = userRepository.query(queryPage, projectId, sourceType)
+
+    return userMapper.fromRestSourceUsers(records, page)
```
blootsvoets: This should probably check for `SUBJECT_READ`. If no projectId is requested, still the individual users should be checked on whether the user has access to those. I wouldn't mind making projectId mandatory though to avoid permission escalation.
Summary:
- add `@NeedsPermission(SUBJECT, READ)` (if possible with projectId)
- add a filter for all users that are mapped: `.filter { auth.checkPermissionOnSubject(SUBJECT, READ, it.projectId, it.login)`

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RestSourceUserResource.kt
```diff
@@ -0,0 +1,154 @@
...
+    @Path("{id}")
+    fun readUser(@PathParam("id") userId: Long): RestSourceUserDTO {
+        val user = ensureUser(userId)
+        auth.checkPermissionOnSubject(Permission.SUBJECT_READ, user.projectId, user.userId)
+        return userMapper.fromEntity(user)
+    }
+
+    @DELETE
+    @Path("{id}")
+    fun deleteUser(@PathParam("id") userId: Long): Response {
```
blootsvoets, line 90: I'm thinking now that before any database request is made, the SUBJECT_READ scope should be checked for being present. Either with a `@NeedsPermission` or with an `auth.checkPermission` call in `ensureUser`. This avoids unauthorised users to make database requests.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/ProjectResource.kt
nivemaham: Yep. You are right.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RestSourceUserResource.kt
nivemaham: Am I right to understand that NeedsPermission only check for permission, but no dependency on subject or project. Then yes. Its better to add the check for permission.

blootsvoets: Indeed, it can only check for project / subject if that is included in the path. But if you leave them out, it just calls `auth.checkPermission`

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RestSourceUserResource.kt
nivemaham: I didn't make the projectId mandatory, since it would make the implementation of the REST-Sounce-Connector more complicated. At the moment, it only requests for users of a particular source-type.

blootsvoets (approved):

PR #69, opened 2020-09-03
Add state-store support

nivemaham:
add state-store check to validate against spoofing


authorizer-app-backend/src/main/java/org/radarbase/authorizer/util/StateStore.kt
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    data class State(val uuid: String, val sourceType: String) {
+        override fun toString() = "state=$uuid&sourceType=$sourceType"
+        companion object {
+            fun toState(string: String) : State {
+                val map = string.split("&")
+                return State(map[0].split("=")[1], map[1].split("=")[1])
+            }
+        }
```
blootsvoets, line 54: Does this work, with additional parameters inside a single parameter?

blootsvoets (approved):
LGTM. Not sure whether joining two parameters to a state works, but if it's tested I'm all for it.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/util/StateStore.kt
nivemaham: I have tested the above implementation. e.g. For the above `toString()`, the `toState()` works fine. If we need to pass additional parameters in the future, maybe we can opt for more complicated solution.

blootsvoets:
@nivemaham additional consideration: why not encode the source type in the path As a path segment, and in the state only pass a UUID, which can also be the ID referencing more parts of the state?

PR #77, opened 2020-10-20
calculate authorized status based on current state

nivemaham:


authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RestSourceUserResource.kt
```diff
@@ -151,13 +151,11 @@ class RestSourceUserResource(
...
-        if (!user.authorized) {
-            throw HttpApplicationException(Response.Status.PROXY_AUTHENTICATION_REQUIRED, "user_unauthorized", "Refresh token for ${user.userId ?: user.externalUserId} is no longer valid.")
-        }
         // refresh token if current token is already expired.
-        if (Instant.now().isAfter(user.expiresAt)) {
+        if (!user.hasValidToken()) {
             return refreshToken(userId, user)
         }
+
         return TokenDTO(user.accessToken, user.expiresAt)
```
blootsvoets, line 159: Good change. For a more kotlinesque response
```suggestion
        return if (user.hasValidToken()) {
            TokenDTO(user.accessToken, user.expiresAt)
        } else {
            refreshToken(userId, user)
        }
```

authorizer-app-backend/src/main/java/org/radarbase/authorizer/api/RestSourceUserMapper.kt
```diff
@@ -25,7 +25,7 @@ class RestSourceUserMapper {
         projectId = user.projectId,
         userId = user.userId,
         sourceId = user.sourceId,
-        isAuthorized = user.authorized,
+        isAuthorized = user.hasValidToken(),
```
blootsvoets, line 28: This is not the intention though. If `isAuthorized` is false, that indicates that the user needs to manually log in again. But if there still is a valid refresh token, no user interaction is needed.

blootsvoets (changes requested):

authorizer-app/src/app/components/rest-source-authorization/update-rest-source-user.component.html
```diff
@@ -77,6 +77,12 @@ <h3 *ngIf="isEditing">Edit details for User</h3>
            id="isAuthorized" [value]= "restSourceUser.isAuthorized? 'Yes':'No'">
   </div>
 
+  <div class="form-group" *ngIf="isEditing">
+    <label for="hasValidToken">Has Valid Token:</label>
```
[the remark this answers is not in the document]
nivemaham: I see, I can remove it. It is indeed an additional information on the token status. In that case we will only have Authorized state, and it will only state if the user currently has a token or not. Not if it is a valid one.

blootsvoets (approved):
LGTM

PR #78, opened 2020-10-28
Handle backends that do not always process multiple states

blootsvoets:
Fixes an issue where a re-login during authorization attempt removes the & symbol.


blootsvoets:
I'm not sure if it's the reason, but quite frequently with the latest version, I get messages like
```
9088802 [grizzly-http-server-0] INFO  - Authorizing with code 8f0b1a0dd797b455e55905523d7c56e058bbd263 state state=NKusZkNl
9088805 [grizzly-http-server-0] ERROR - [500] /rest-sources/backend/users
java.lang.IndexOutOfBoundsException: Index: 1, Size: 1
	at java.base/java.util.Collections$SingletonList.get(Collections.java:4847)
	at org.radarbase.authorizer.util.StateStore$State$Companion.toState(StateStore.kt:61)
```

nivemaham:
My guess is that this StateStore implementation makes the component stateful. Thus maybe the pods don't have identical state. I have not seen this issue when I was running with a single pod. If you are still seeing this with, a single pod then I am wrong.

PR #81, opened 2020-10-29
Allow overriding existing user if it does not have a project

### LINKED ISSUE #80: User without a project cannot be created multiple times
As long as the project ID of a newly authorised user hasn't been set, it cannot be found anywhere else in the interface. However, it also cannot be re-authorized. There should be a rule that allows overriding a user which has a `null` project ID without creating a conflict.

blootsvoets:
Fixes #80.


nivemaham (approved):
LGTM

PR #88, opened 2020-11-18
Add human readable

peyman-mohtashami:
Changes:

- Remove edit button from table
- Disable editable fields in info page
- Update userId with externalId if available

I checked all the pages except register page. It could be possible after deployment.


blootsvoets:
Looks good, I'll verify in production. Could you add the registration form details as well? E.g. dropdown with `externalId ? externalId : userId`?

peyman-mohtashami:
> Looks good, I'll verify in production. Could you add the registration form details as well? E.g. dropdown with `externalId ? externalId : userId`?

This is actually not in 'register' component. The registration form is in update form, and it's been added already. I just need to test it.

authorizer-app/src/app/components/rest-source-authorization/rest-source-user-list.component.html
```diff
@@ -40,11 +40,11 @@
...
       <ng-container matColumnDef="userId">
         <th mat-header-cell *matHeaderCellDef mat-sort-header> User ID </th>
-        <td mat-cell *matCellDef="let user"> {{user.userId |  slice:0:8 }}... </td>
+        <td mat-cell *matCellDef="let user"> {{user.externalId? user.externalId : ((user.userId | slice:0:8) + '...')}}</td>
       </ng-container>
       <ng-container matColumnDef="externalUserId">
         <th mat-header-cell *matHeaderCellDef mat-sort-header> External User ID </th>
-        <td mat-cell *matCellDef="let user"> {{user.externalUserId}} </td>
+        <td mat-cell *matCellDef="let user"> {{user.serviceUserId}} </td>
       </ng-container>
```
blootsvoets, line 48: Perhaps the external user ID does not even need to be shown here, instead the first header could be the actual user ID and the second column the external user ID.

authorizer-app/src/app/components/rest-source-authorization/rest-source-user-list-reset-dialog.html
```diff
@@ -1,6 +1,6 @@
 <h1 mat-dialog-title>Reset User in project {{data.projectId}}?</h1>
 <div mat-dialog-content>
-  <p>This will pull all the data for the user (project: {{data.projectId}}, user-id: {{data.userId}} with external-user-id: {{data.externalUserId}}) again in the specified date range.</p>
+  <p>This will pull all the data for the user (project: {{data.projectId}}, user-id: {{data.externalId? data.externalId : data.userId}} with external-user-id: {{data.serviceUserId}}) again in the specified date range.</p>
```
blootsvoets, line 3: If the external user ID is shown, it would still be helpful to see the actual user ID between brackets, i.e. `blootsvoets (abcde-aefa)`

authorizer-app/src/app/components/rest-source-authorization/rest-source-user-list-delete-dialog.html
```diff
@@ -1,7 +1,8 @@
 <h1 mat-dialog-title>WARNING!</h1>
 <div mat-dialog-content>
   <p>Are you sure you want to delete the user with- <strong>id:</strong> {{data.id}},
-    <strong>user-id:</strong> {{data.userId}} and <strong>project-id:</strong> {{data.projectId}}?
+    <strong>user-id:</strong> {{data.externalId? data.externalId : data.userId}}
```
blootsvoets, line 4: If the external user ID is shown, it would still be helpful to see the actual user ID between brackets, i.e. `blootsvoets (abcde-aefa)`

authorizer-app/src/app/components/rest-source-authorization/update-rest-source-user.component.html
```diff
@@ -9,11 +9,11 @@
 
 <div *ngIf="restSourceUser" class="container col-md-4">
   <h3 *ngIf="!isEditing">Enter details for User</h3>
-  <h3 *ngIf="isEditing">Edit details for User</h3>
+  <h3 *ngIf="isEditing">Details for User</h3>
   <div class="form-group">
     <label for="externalUserId">External User ID:</label>
```
blootsvoets, line 14: Perhaps rename to (REST?) Service user ID

authorizer-app/src/app/components/rest-source-authorization/update-rest-source-user.component.html
```diff
@@ -33,9 +33,9 @@ <h3 *ngIf="isEditing">Edit details for User</h3>
 
   <div class="form-group" *ngIf="restSourceUser.projectId">
     <label for="userId">User ID:</label>
-    <select class="form-control" id="userId" [(ngModel)]="restSourceUser.userId" [disabled]="isEditing">
+    <select class="form-control" id="userId" [(ngModel)]="restSourceUser.externalId ? restSourceUser.externalId : restSourceUser.userId" #userId="ngModel" [disabled]="isEditing">
       <option [selected]="true" value="">Select a user</option>
-      <option [value]="user.id" *ngFor="let user of subjects">{{user.id}}</option>
+      <option *ngFor="let user of subjects" [value]="user.externalId ? user.externalId : user.id">{{user.externalId? user.externalId : user.id}}</option>
```
blootsvoets, line 38: Note that the value of an option should always be the `userId`, since the backend only processes that. Preferably, this list is sorted in the following way

1. items with external ID, sorted by externalId, ASC, perhaps again with displayed as `externalId (userId | slice:0:8)`
2. items without external ID, sorted by userId, ASC

To make this a bit more user friendly, could this a combobox, dropdown with text box instead? https://material.angular.io/components/autocomplete/overview
If possible the filter would select both from userId and externalId.

blootsvoets (commented):
I've added some comments to make this a bit more interoperable with using plain user ID's in the interface.

authorizer-app/src/app/components/rest-source-authorization/rest-source-user-list.component.html
peyman-mohtashami: I'm confused! We have:
1) `userId`
2) `externalId`: (was `HumanReadableUserId` before)
3) `serviceUserId`: (was `externalUserId` before)
You mean we don't need `serviceUserId` in the table?

authorizer-app/src/app/components/rest-source-authorization/update-rest-source-user.component.html
peyman-mohtashami: we used `external-user-id` or `External User ID` in many places. You mean change them to `service-user-id` or `Service User ID`?

blootsvoets: I think we don't need the serviceUserId in the table, as long as users can see it in the info dialog.

authorizer-app/src/app/components/rest-source-authorization/rest-source-user-list.component.html
blootsvoets: I think we don't need the `serviceUserId` in the table, as long as users can see it in the info dialog.

peyman-mohtashami:
@blootsvoets I tried to add autocomplete, but it needs some efforts for validation of input field which should be restricted to values from dropdown.
I left it as a simple dropdown for now. If you think it is necessary, let me know to continue.

peyman-mohtashami:
Fix "Access denied" redirection
Fix "user" and "project" form select for null values.

@blootsvoets Please review last three commits.

blootsvoets (approved):
LGTM

mpgxvii:
Hi @peyman-mohtashami @blootsvoets  just wondering, was there a specific reasons to disable the editable fields here? I think some of the studies require some fields to remain editable.

blootsvoets:
@mpgxvii iirc, the only change is that the dates are now not editable. The reason was that editing these fields would not reset the user version ID and data collection, so the rest connector would not consistently use the new dates. In the reset dialog, the dates can still be edited and the new dates are also reflected in data collection.

PR #93, opened 2020-12-15
Check projectId before request

peyman-mohtashami:


blootsvoets:
It turns out this was the offending line of code in our other code base, but the solution was the previous value
`if (projectId)`. So this PR identifies the issue but is not needed to solve it.

PR #90, opened 2020-11-26
Add task for sending a deregister request to Garmin

mpgxvii:
- Add timer task for sending a deregister request every hour to Garmin for users with elapsed end date (older than 14 days)

PR #94, opened 2021-01-11
Add RestSourceUser createdAt property 

mpgxvii:
- Adds RestSourceUser `createdAt` property for determining the end date of the backfill (for sources that uses backfill) or up to what date the historic collection must be executed

Depends on #90 (part of providing Garmin support)
https://github.com/RADAR-base/RADAR-PushEndpoint/pull/6


authorizer-app-backend/src/main/java/org/radarbase/authorizer/doa/entity/RestSourceUser.kt
```diff
@@ -29,6 +29,10 @@ class RestSourceUser {
     @SequenceGenerator(name = "sequenceGenerator", sequenceName = "rest_source_user_id_seq", initialValue = 1, allocationSize = 1)
     var id: Long? = null
 
+    // Date when the user is created
+    @Column(name = "created_at")
+    var createdAt: Instant = Instant.now()
```
yatharthranjan, line 34: Since nullable constraint is set to true in DB, you should make this var nullable too.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/api/ApiDeclarations.kt
```diff
@@ -67,6 +67,7 @@ data class ShareableClientDetails(
 
 class RestSourceUserDTO(
     val id: String?,
+    val createdAt: Instant,
```
yatharthranjan, line 70: Similarly, this can be null.

authorizer-app-backend/src/main/resources/db/changelog/changes/00000000000003_update_schema.xml
```diff
@@ -0,0 +1,16 @@
...
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd
+                        http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
+
+    <changeSet author="pauline" id="addCreatedAt">
+        <addColumn tableName="rest_source_user">
+            <column name="created_at" type="timestamp" defaultValueComputed="CURRENT_TIMESTAMP">
+                <constraints nullable="true" />
```
yatharthranjan, line 12: What happens to users which don't have computed values (prior registered users). I know this will probably never happen (since we will deploy after this is merged) but would be good to handle the case. I think those should have null values. AFAIR this is specified by the `value=NULL` property which will set existing rows to this value for the new column.

yatharthranjan (commented):
Thanks for the changes. Please see the comments below.

authorizer-app-backend/src/main/resources/db/changelog/changes/00000000000003_update_schema.xml
mpgxvii: I was thinking of making the default value `defaultValueComputed="CURRENT_TIMESTAMP"`?

yatharthranjan: But that will only take effect on newly created users. It will not update the column values existing users (if any already present). Also even if it does, using the current timestamp for existing users create date would be incorrect since they would have authorised in the past. If it is null the external system can handle this case.

mpgxvii: Oh okay I see. I thought that the external system needed an actual value. I'll make the changes.

yatharthranjan: hi Sorry, i think there was a misunderstanding. The new records should have the computed current timestamp (so keeping `defaultValueComputed="CURRENT_TIMESTAMP"` as it is) but also adding value=null for any existing records. This will only update the existing records without affecting the default value of the column.

mpgxvii: Ah, my bad I've made the changes.

authorizer-app-backend/src/main/resources/db/changelog/changes/db.changelog-master.xml
```diff
@@ -7,4 +7,5 @@
   <include  file="db/changelog/changes/00000000000000_initial_schema.xml"/>
   <include  file="db/changelog/changes/00000000000001_update_schema.xml"/>
   <include  file="db/changelog/changes/00000000000002_update_schema.xml"/>
+  <include  file="db/changelog/changes/00000000000003_update_schema.xml"/>
```
blootsvoets, line 10: I see the same thing has happened in previous files, but could you please rename the file to `00000000000003_add_created_at_column.xml`? That way if we ever need to troubleshoot the table layout it's easier to locate the issue.

blootsvoets (approved):
LGTM

PR #97, opened 2021-01-27
Add endpoint for signing OAuth1 requests

mpgxvii:
- Add endpoint for signing OAuth1 requests (instead of sending the tokenSecret/refresh token in the token endpoint)
- Parameters required are `SignRequestParams`: `{url: "", method: "", parameters: {}}`

### DIFF
--- a/authorizer-app-backend/src/main/java/org/radarbase/authorizer/api/ApiDeclarations.kt
+++ b/authorizer-app-backend/src/main/java/org/radarbase/authorizer/api/ApiDeclarations.kt
@@ -41,6 +41,12 @@ data class RestOauth1UserId(
     @JsonProperty("userId") var userId: String,
 )
 
+data class SignRequestParams(
+    var url: String,
+    var method: String,
+    val params: MutableMap<String, String?>
+)
+
 data class RequestTokenPayload(
     var sourceType: String,
     var code: String? = null,
@@ -94,8 +100,12 @@ data class RestSourceUsers(
 
 class TokenDTO(
     val accessToken: String?,
-    val expiresAt: Instant?,
-    val refreshToken: String?,
+    val expiresAt: Instant?
+)
+
+class UrlSignatureDTO(
+    val url: String,
+    val signedUrl: String?
 )
 
 data class Page(
--- a/authorizer-app-backend/src/main/java/org/radarbase/authorizer/api/ApiDeclarations.kt
+++ b/authorizer-app-backend/src/main/java/org/radarbase/authorizer/api/ApiDeclarations.kt
@@ -44,7 +44,7 @@ data class RestOauth1UserId(
 data class SignRequestParams(
     var url: String,
     var method: String,
-    val params: Map<String, String?>,
+    val parameters: Map<String, String?>,
 )
 
 data class RequestTokenPayload(
@@ -101,11 +101,6 @@ data class RestSourceUsers(
 data class TokenDTO(
     val accessToken: String?,
     val expiresAt: Instant?,
-)
-
-data class UrlSignatureDTO(
-    val url: String,
-    val signedUrl: String?,
 )
 
 data class Page(
--- a/authorizer-app-backend/src/main/java/org/radarbase/authorizer/api/ApiDeclarations.kt
+++ b/authorizer-app-backend/src/main/java/org/radarbase/authorizer/api/ApiDeclarations.kt
@@ -44,7 +44,7 @@ data class RestOauth1UserId(
 data class SignRequestParams(
     var url: String,
     var method: String,
-    val params: Map<String, String?>
+    val params: Map<String, String?>,
 )
 
 data class RequestTokenPayload(
@@ -98,14 +98,14 @@ data class RestSourceUsers(
     val users: List<RestSourceUserDTO>,
 )
 
-class TokenDTO(
+data class TokenDTO(
     val accessToken: String?,
-    val expiresAt: Instant?
+    val expiresAt: Instant?,
 )
 
-class UrlSignatureDTO(
+data class UrlSignatureDTO(
     val url: String,
-    val signedUrl: String?
+    val signedUrl: String?,
 )
 
 data class Page(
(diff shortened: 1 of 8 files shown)


yatharthranjan (approved):

blootsvoets:
Hmm the intention of only sending the access token is that then the token is not refreshed in two places simultaneously, plus credentials are not needlessly delegated. Why not just request a new access token when it is needed for the Garmin backfill?

yatharthranjan:
>Hmm the intention of only sending the access token is that then the token is not refreshed in two places simultaneously, plus credentials are not needlessly delegated. Why not just request a new access token when it is needed for the Garmin backfill?

Access tokens in garmin are fixed. The user will need to de register and then authorise again to get a new access token. There is no other way to refresh the token.

mpgxvii:
Ok yes I can add a new endpoint

nivemaham:
> > Hmm the intention of only sending the access token is that then the token is not refreshed in two places simultaneously, plus credentials are not needlessly delegated. Why not just request a new access token when it is needed for the Garmin backfill?
> 
> Access tokens in garmin are fixed. The user will need to de register and then authorise again to get a new access token. There is no other way to refresh the token.

Then why do we need the refresh_token with the access_token?. If user has to reauthorise, wouldn't it be better to use the PUT operation to update the access token in the back-end?

blootsvoets:
Alternatively, if you need to refresh the same access token, could the rest source authorizer return the same access token (since it is fixed) but refresh it when needed?

yatharthranjan:
>Then why do we need the refresh_token with the access_token?. If user has to reauthorise, wouldn't it be better to use the PUT operation to update the access token in the back-end?

The user doesn't have to reauthorise. I was just giving the only way to get new access token. We don't need the new access token as the one will be valid.


> Alternatively, if you need to refresh the same access token, could the rest source authorizer return the same access token (since it is fixed) but refresh it when needed?
so
GET token
(Refresh if needed, return access token)
POST token/refresh
(Refresh, return access token)

We need the User Access token Secret (which is considered refresh token here) for signing the OAuth requests for backfill. Both the access token and access token secret are required.

blootsvoets:
Ah in that case, it might be better indeed to have a new endpoint that can sign a request:
POST token/sign
Param1=a&param2=b
===
{
Url: ...
SignedUrl: ...
}

yatharthranjan:
Yes, that sounds good to me. @mpgxvii can you make the suggested change (i think most of the signing code is already there) and I will update the code in push endpoint repo.

mpgxvii:
> Ah in that case, it might be better indeed to have a new endpoint that can sign a request:
> 
> ```
> POST token/sign
> {url: “...?Param1=a&param2=b”}
> ===
> {
> Url: ...
> SignedUrl: ...
> }
> ```



> Yes, that sounds good to me. @mpgxvii can you make the suggested change (i think most of the signing code is already there) and I will update the code in push endpoint repo.

Okay sure I can make the changes.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OAuth2RestSourceAuthorizationService.kt
```diff
@@ -102,7 +102,11 @@ class OAuth2RestSourceAuthorizationService(
         throw HttpBadRequestException("", "Not available for auth type")
     }
 
-    private fun post(form: FormBody, sourceType: String): Request {
+    override fun signUrl(user: RestSourceUser, url: String, method: String, params: MutableMap<String, String?>): String {
+        return ""
```
blootsvoets, line 106: perhaps this could throw a `HttpBadRequestException`, since this authorization service does not support signing a URL? Or a `HttpNotFoundException` since the `sign` url does not exist for a OAuth 2.0 user?

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OAuth2RestSourceAuthorizationService.kt
```diff
@@ -102,7 +102,11 @@ class OAuth2RestSourceAuthorizationService(
...
+    }
+
+        private fun post(form: FormBody, sourceType: String): Request {
```
blootsvoets, line 109: Indentation was increased here.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OAuth1RestSourceAuthorizationService.kt
```diff
@@ -146,6 +146,15 @@ abstract class OAuth1RestSourceAuthorizationService(
             .build()
     }
 
+    override fun signUrl(user: RestSourceUser, url: String, method: String, params: MutableMap<String, String?>): String {
+        val authConfig = configMap[user.sourceType]
+            ?: throw HttpBadRequestException("client-config-not-found",
+                "Cannot find client configurations for source-type ${user.sourceType}")
+        val tokens = RestOauth1AccessToken(user.accessToken!!, user.refreshToken)
```
blootsvoets, line 153: Instead of asserting an access token exists with `!!`, it would be better to check that it exists or throw a proper HTTP exception.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/api/ApiDeclarations.kt
```diff
@@ -41,6 +41,12 @@ data class RestOauth1UserId(
     @JsonProperty("userId") var userId: String,
 )
 
+data class SignRequestParams(
+    var url: String,
+    var method: String,
+    val params: MutableMap<String, String?>
```
blootsvoets, line 47: It seems that a `MutableMap` is not needed, a `Map` would suffice. This logic extends to places where this class is used.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/api/ApiDeclarations.kt
```diff
@@ -94,7 +100,12 @@ data class RestSourceUsers(
 
 class TokenDTO(
     val accessToken: String?,
-    val expiresAt: Instant?,
+    val expiresAt: Instant?
+)
+
+class UrlSignatureDTO(
```
blootsvoets, line 106: Please use data class for this and the above TokenDTO

blootsvoets (approved):
One small comment, good to merge otherwise.

blootsvoets (approved):
LGTM

PR #98, opened 2021-02-03
Expose endpoint for revoking/deregistering access tokens

mpgxvii:
- Exposes endpoint for revoking/deregistering access tokens needed for deregistering users in the PushEndpoint https://github.com/RADAR-base/RADAR-PushEndpoint/pull/11 at `DELETE source-clients/{type}/authorization/{serviceUserId}`
- Adds endpoint for getting user by serviceUserId at `GET source-clients/{type}/authorization/{serviceUserId}`
- Separate user deletion implementations for different auth types (so that Garmin auth can immediately delete while Fitbit auth can revoke first before deleting)
- Change Garmin deregistration endpoint to immediately delete the user 

Partially solves https://github.com/RADAR-base/RADAR-PushEndpoint/issues/12

### DIFF
--- a/authorizer-app-backend/src/main/java/org/radarbase/authorizer/doa/RestSourceUserRepositoryImpl.kt
+++ b/authorizer-app-backend/src/main/java/org/radarbase/authorizer/doa/RestSourceUserRepositoryImpl.kt
@@ -168,6 +168,24 @@ class RestSourceUserRepositoryImpl(
         }
     }
 
+    override fun findByExternalId(
+        externalId: String,
+        sourceType: String,
+    ): RestSourceUser {
+        var queryString = """
+               SELECT u
+               FROM RestSourceUser u
+               WHERE u.externalId < :externalId
+               AND u.sourceType = :sourceType
+        """.trimIndent()
+        return transact {
+            val query = createQuery(queryString, RestSourceUser::class.java)
+            query.setParameter("sourceType", sourceType)
+            query.setParameter("externalId", externalId)
+            query.singleResult
+        }
+    }
+
     override fun delete(user: RestSourceUser) = transact {
         remove(merge(user))
     }
--- a/authorizer-app-backend/src/main/java/org/radarbase/authorizer/doa/RestSourceUserRepositoryImpl.kt
+++ b/authorizer-app-backend/src/main/java/org/radarbase/authorizer/doa/RestSourceUserRepositoryImpl.kt
@@ -171,19 +171,20 @@ class RestSourceUserRepositoryImpl(
     override fun findByExternalId(
         externalId: String,
         sourceType: String,
-    ): RestSourceUser {
+    ): RestSourceUser? {
         var queryString = """
                SELECT u
                FROM RestSourceUser u
-               WHERE u.externalId = :externalId
+               WHERE u.externalUserId = :externalId
                AND u.sourceType = :sourceType
         """.trimIndent()
-        return transact {
+        val result = transact {
             val query = createQuery(queryString, RestSourceUser::class.java)
             query.setParameter("sourceType", sourceType)
             query.setParameter("externalId", externalId)
-            query.singleResult
+            query.resultList
         }
+        return if (result.isEmpty()) null else result.get(0)
     }
 
     override fun delete(user: RestSourceUser) = transact {
--- a/authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RestSourceUserResource.kt
+++ b/authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RestSourceUserResource.kt
@@ -21,7 +21,6 @@ import org.radarbase.authorizer.api.*
 import org.radarbase.authorizer.doa.RestSourceUserRepository
 import org.radarbase.authorizer.doa.entity.RestSourceUser
 import org.radarbase.authorizer.service.RestSourceAuthorizationService
-import org.radarbase.authorizer.util.OauthSignature
 import org.radarbase.authorizer.util.StateStore
 import org.radarbase.jersey.auth.Auth
 import org.radarbase.jersey.auth.Authenticated
@@ -142,6 +141,23 @@ class RestSourceUserResource(
         }
         userRepository.delete(user)
         return Response.noContent().header("user-removed", userId).build()
+    }
+
+    @POST
+    @Path("revoke/sourceType/{sourceType}/externalId/{externalId}")
+    @NeedsPermission(Permission.Entity.MEASUREMENT, Permission.Operation.READ)
+    fun revokeToken(@PathParam("externalId") externalId: String,
+                    @PathParam("sourceType") sourceType: String,
+                    accessToken: TokenDTO): Boolean {
+        val user = userRepository.findByExternalId(externalId, sourceType)
+        if (user == null) {
+            logger.info("No user found for external ID provided. Continuing deregistration..")
+            return authorizationService.revokeToken(externalId, sourceType)
+        }
+        else {
+            auth.checkPermissionOnSubject(Permission.MEASUREMENT_READ, user.projectId, user.userId)
+            return authorizationService.revokeToken(user)
+        }
     }
 
     @POST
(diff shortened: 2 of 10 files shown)


authorizer-app-backend/src/main/java/org/radarbase/authorizer/doa/RestSourceUserRepositoryImpl.kt
```diff
@@ -168,6 +168,24 @@ class RestSourceUserRepositoryImpl(
...
     }
 
+    override fun findByExternalId(
+        externalId: String,
+        sourceType: String,
+    ): RestSourceUser {
+        var queryString = """
+               SELECT u
+               FROM RestSourceUser u
+               WHERE u.externalId < :externalId
```
blootsvoets, line 178:
```suggestion
               WHERE u.externalId = :externalId
```

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RestSourceUserResource.kt
```diff
@@ -144,6 +143,23 @@ class RestSourceUserResource(
         return Response.noContent().header("user-removed", userId).build()
     }
 
+    @POST
+    @Path("revoke/sourceType/{sourceType}/externalId/{externalId}")
```
blootsvoets, line 147: Alternative to match other api: `DELETE source-clients/{type}/authorization/{serviceUserId}` and move it to the source-clients resource. It might be useful to add a get method there as well, to get the user info and perhaps the token of a given serviceUserId.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RestSourceUserResource.kt
```diff
@@ -144,6 +143,23 @@ class RestSourceUserResource(
...
+    @NeedsPermission(Permission.Entity.MEASUREMENT, Permission.Operation.READ)
+    fun revokeToken(@PathParam("externalId") externalId: String,
+                    @PathParam("sourceType") sourceType: String,
+                    accessToken: TokenDTO): Boolean {
```
blootsvoets, line 151: Access token does not seem to be used. In any case, this service contains all access tokens, so it wouldn’t need the client to pass it.

mpgxvii: My bad the `revokeToken` method should have included the accessToken in the parameters. I was thinking that in the case where the user is not part of the repository, we can still revoke the token (using the accessToken as a request parameter) so that it won't keep sending data when it was not authorized (in the rest sources auth repo) in the first place?

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/SourceClientResource.kt
```diff
@@ -76,6 +77,35 @@ class SourceClientResource(
         return authorizationService.getAuthorizationEndpointWithParams(type, callbackUrl);
     }
 
+    @DELETE
+    @Path("{type}/authorization/{serviceUserId}")
+    @NeedsPermission(Permission.Entity.MEASUREMENT, Permission.Operation.READ)
+    fun deleteAuthorization(@PathParam("serviceUserId") serviceUserId: String,
```
blootsvoets, line 83: I’m not sure if `DELETE` takes a body. Could the access token be passed as an (optional) query param?

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RestSourceUserResource.kt
```diff
@@ -199,9 +197,10 @@ class RestSourceUserResource(
     @NeedsPermission(Permission.Entity.SUBJECT, Permission.Operation.UPDATE)
     fun reportDeregistration(
         @PathParam("id") userId: Long,
-    ): RestSourceUserDTO {
-        val existingUser = ensureUser(userId)
-        return userMapper.fromEntity(authorizationService.deRegisterUser(existingUser))
+    ): Response {
+        val user = ensureUser(userId)
+        authorizationService.deRegisterUser(user)
+        return Response.noContent().header("user-removed", userId).build()
```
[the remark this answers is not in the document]
yatharthranjan: I think this endpoint can also move to SourceClientResource with path `POST /{source-type}/deregister/`, then we can configure `/garmin/deregister` endpoint on the garmin portal.

yatharthranjan (commented):
Please find some comments below.

Also, was this tested with the garmin portal?

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OAuth1RestSourceAuthorizationService.kt
```diff
@@ -86,11 +86,32 @@ abstract class OAuth1RestSourceAuthorizationService(
...
+            }
+        }
+    }
+
+    override fun deRegisterUser(user: RestSourceUser) {
+        return userRepository.delete(user)
+    }
+
+    override fun deleteUser(user: RestSourceUser) {
+        logger.info("Deleting user...")
```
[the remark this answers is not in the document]
mpgxvii: I think I mentioned it before about adding a Revoke Token button in the front end. What if we only allow deletions when the user is already unauthorized? So the deregister endpoint will only mark as unauthorized and the user will have to manually delete the user in the front end?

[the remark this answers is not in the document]
yatharthranjan: I guess we should try not to complicate this for the users, we can just revokeToken and immediately delete the user after that. Garmin will send a deregister request to our endpoint later but the user will not be present in the repo so we will just return 200. This will be fine I think?

Worst case the user was not revoked properly by Garmin, then it will send some further data for that user to push endpoint, which will then again find that user is unauthorized (since not in repo) and revoke token again, so trying the flow again.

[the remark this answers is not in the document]
mpgxvii: Oh ok I see. Thanks

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/SourceClientResource.kt
```diff
@@ -107,8 +107,11 @@ class SourceClientResource(
         @PathParam("type") sourceType: String,
     ): RestSourceUserDTO {
         val user = userRepository.findByExternalId(serviceUserId, sourceType)
-        auth.checkPermissionOnSubject(Permission.MEASUREMENT_READ, user.projectId, user.userId)
-        return userMapper.fromEntity(user)
+        if (user != null) {
+            auth.checkPermissionOnSubject(Permission.MEASUREMENT_READ, user.projectId, user.userId)
+            return userMapper.fromEntity(user)
+        }
+        else throw HttpNotFoundException("user-not-found", "User with service user id not found.")
```
blootsvoets, line 114:
```suggestion
            ?: throw HttpNotFoundException("user-not-found", "User with service user id not found.")

        auth.checkPermissionOnSubject(Permission.MEASUREMENT_READ, user.projectId, user.userId)
        return userMapper.fromEntity(user)
```

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/SourceClientResource.kt
```diff
@@ -120,13 +119,13 @@ class SourceClientResource(
...
     @POST
     @Path("{type}/deregister")
-    fun reportDeregistration(@PathParam("type") sourceType: String, params: DeregistrationParams): Response {
-        val user = userRepository.findByExternalId(params.userId, sourceType)
-        if (user != null) {
-            if (user.accessToken == params.userAccessToken)
+    fun reportDeregistration(@PathParam("type") sourceType: String, body: DeregistrationsDTO): Response {
+        body.deregistrations.forEach { it ->
+            val user = userRepository.findByExternalId(it.userId, sourceType)
+            if (user != null && user.accessToken == it.userAccessToken)
```
yatharthranjan, line 125: Right, at least it's consistent in the format and sending data for multiple users 😆

yatharthranjan (dismissed):
LGTM, some further comments below.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OAuth2RestSourceAuthorizationService.kt
```diff
@@ -84,6 +85,9 @@ class OAuth2RestSourceAuthorizationService(
         return httpClient.request(post(form, user.sourceType))
     }
 
+    override fun revokeToken(externalId: String, sourceType: String, token: String): Boolean =
+        throw HttpBadRequestException("", "Not available for auth type")
```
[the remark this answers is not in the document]
mpgxvii: Ok I'll add it in another PR

Issue #102, opened 2021-02-18
Pagination issue (only displaying 100 users)


yatharthranjan:
After migrating from 1.x.x to 2.0, All the users are not displayed, only 100 users are displayed.

yatharthranjan:
fixed in 3.0.0

PR #104, opened 2021-02-22
Fix user list pagination

mpgxvii:
- Fix user list pagination by adding `metadata` property to `RestSourceUsers` to obtain page data and by fixing front end 

Solves #102


authorizer-app/src/app/components/rest-source-authorization/rest-source-user-list.component.html
```diff
@@ -82,8 +82,10 @@
     </table>
 
     <mat-paginator
+      [length]="totalItems"
       [pageSize]="20"
-      [pageSizeOptions]="[5, 10, 20, 50]"
+      [pageSizeOptions]="[1, 5, 10, 20, 50]"
```
blootsvoets, line 87: options 1 and 5 can be removed?

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RestSourceUserResource.kt
```diff
@@ -64,7 +64,7 @@ class RestSourceUserResource(
         @QueryParam("source-type") sourceType: String?,
         @QueryParam("size") pageSize: Int?,
         @DefaultValue("1") @QueryParam("page") pageNumber: Int,
-    ): RestSourceUsers {
```
blootsvoets, line 67: Why not just add a `metadata` element to restsourceusers, including page, page size and total items? If we need metadata in another element, we can reuse the metadata type.

blootsvoets (commented):
Looks OK, though I’d prefer putting the metadata in JSON to be explicit about it. In any case, using the `X-` header prefix is no longer recommended.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RestSourceUserResource.kt
```diff
@@ -74,7 +74,7 @@ class RestSourceUserResource(
                 .map { it.id }
         }
 
-        if (projects.isEmpty()) return RestSourceUsers(emptyList())
+        if (projects.isEmpty()) return RestSourceUsers(emptyList(), Page())
```
blootsvoets, line 77: for completeness, you can set
```suggestion
        if (projects.isEmpty()) return RestSourceUsers(emptyList(), Page(totalElements = 0))
```

blootsvoets (approved):
LGTM

mpgxvii:
@blootsvoets Sorry would you be able to approve again? The latest commit dismissed the review.

blootsvoets (approved):

Issue #109, opened 2021-02-25
Display RESET version info in the Subject Info Page


yatharthranjan:
It would be useful to expose the Reset version info in the subject info page as it gives an indication of when the subject was last reset.

yatharthranjan:
Done

Issue #108, opened 2021-02-25
Whitespace in the Project Name results in 404


yatharthranjan:
If there is a whitespace in the Project name, then get users fails with 404. This is supported in MP by URL encoding.

yatharthranjan:
Fixed

Issue #107, opened 2021-02-25
User UUID display


yatharthranjan:
Currently, the User UUID is truncated a lot. It would be nice to give that column some more space since other columns (like Authorized, External User Id and Actions columns) have a lot of empty space. Otherwise, at least it should show the full UUID when the cursor is hovering on the ID. The same should be shown on the Info page of the subject.

yatharthranjan:
fixed

PR #111, opened 2021-02-26
Minor fixes 

mpgxvii:
- Show reset version info in user info page (or `update-rest-source-user` component) (Solves #109)
- Encode uri when getting projects to allow project ids to have whitespaces (Solves #108)
- Show complete user id in `update-rest-source-user` component and show a tooltip with the complete user id in the user list component (Solves #107)


blootsvoets (dismissed):
LGTM. Part of our idea to have only 8 characters is that it’s already quite likely unique and easier to compare (like git does).  But 13 is reasonable as well.

blootsvoets (approved):
LGTM. Would there be a way to wrap encodeURI/encodeURIComponent around all project ID instances? Or are we confident we already have everything.

mpgxvii:
> LGTM. Would there be a way to wrap encodeURI/encodeURIComponent around all project ID instances? Or are we confident we already have everything.

Yes those 2 are the only ones using `projectId`.

PR #114, opened 2021-03-04
Mark user as unauthorized after revoking token

mpgxvii:
- Mark user as unauthorized after revoking token

### DIFF
--- a/authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OAuth1RestSourceAuthorizationService.kt
+++ b/authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OAuth1RestSourceAuthorizationService.kt
@@ -69,6 +69,7 @@ abstract class OAuth1RestSourceAuthorizationService(
     }
 
     override fun revokeToken(user: RestSourceUser): Boolean {
+        val userId = user.id ?: return false
         val authConfig = configMap[user.sourceType]
             ?: throw HttpBadRequestException("client-config-not-found",
                 "Cannot find client configurations for source-type ${user.sourceType}")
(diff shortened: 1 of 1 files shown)


authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OAuth1RestSourceAuthorizationService.kt
```diff
@@ -69,17 +69,21 @@ abstract class OAuth1RestSourceAuthorizationService(
...
         val req = createRequest("DELETE",
             authConfig.deregistrationEndpoint!!,
-            RestOauth1AccessToken(user.accessToken!!, user.refreshToken),
+            RestOauth1AccessToken(accessToken, user.refreshToken),
             user.sourceType)
 
         return httpClient.newCall(req).execute().use { response ->
             when (response.code) {
-                200, 204 -> true
+                200, 204 -> !this.userRepository.updateToken(null, userId).authorized
```
yatharthranjan, line 86:
```suggestion
                200, 204 -> {
                    this.userRepository.updateToken(null, userId)
                    true
                  }
```
Just to make it clear.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OAuth1RestSourceAuthorizationService.kt
```diff
@@ -69,17 +69,21 @@ abstract class OAuth1RestSourceAuthorizationService(
     }
 
     override fun revokeToken(user: RestSourceUser): Boolean {
+        val userId = user.id ?: return false
+        val accessToken = user.accessToken
+        if (accessToken == null || !user.authorized) return false
```
yatharthranjan, line 74: should these checks throw an exception?

yatharthranjan (approved):
LGTM

Issue #96, opened 2021-01-19
Add CI


yatharthranjan:
At least minimal CI should be added via travis or github actions that check the build is successful. Further, tests can be added for even better CI.

mpgxvii:
Solved in #127

PR #127, opened 2021-05-04
Add Github actions CI

mpgxvii:
- Add initial Github actions CI

Solves #96


blootsvoets (dismissed):
Looks good to me. Is the npm run test independent from the Java code? If so, you could separate the npm build job from the Java build job. This makes it easier to see what part of the build fails and makes the cache more reusable.

blootsvoets:
For reference, see https://github.com/RADAR-base/radar-app-config/blob/master/.github/workflows/main.yml

blootsvoets (approved):
Thanks for the update, looks very good.

PR #130, opened 2021-05-25
Release 3.1.1

blootsvoets:
Changes since version 3.1.0:

- Introduced Github Actions builds
- Backend CLI supports the `--help` flag
- Dependency updates (includes fix for ManagementPortal project retrieval exception).
- Logging improvements
- Fixed selecting a source type in Safari if only one source type is available.
- Add client-side caching


authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/ProjectResource.kt
```diff
@@ -41,6 +42,7 @@ class ProjectResource(
 
     @GET
     @NeedsPermission(Permission.Entity.PROJECT, Permission.Operation.READ)
+    @Cache(maxAge = 300, isPrivate = true)
```
yatharthranjan, line 45: what is the `isPrivate` property?

yatharthranjan (approved):
LGTM

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/ProjectResource.kt
blootsvoets: See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

 > The response may be stored only by a browser's cache, even if the response is normally non-cacheable. If you mean to not store the response in any cache, use no-store instead. This directive is not effective in preventing caches from storing your response.

I’ve understood this directive as: do not cache by proxy servers or on the server side, since the response is authenticated.

blootsvoets: From the w3c spec https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.9.1:

> private
>> Indicates that all or part of the response message is intended for a single user and MUST NOT be cached by a shared cache. This allows an origin server to state that the specified parts of the response are intended for only one user and are not a valid response for requests by other users. A private (non-shared) cache MAY cache the response.

yatharthranjan: Thanks, that makes sense.

PR #147, opened 2021-07-21
Updated authorization flow

blootsvoets:
Updated authorization flow. Intention is that both project admins and subjects are able to register their REST account with the rest source authorizer. A setup with a token state is used to allow state to be persisted between starting registration and finishing it. If it turns out that this is not feasible, the state could be persisted in the session storage instead.

- [x] created API
- [x] frontend flow
  - [x] add Link new account button on the main page.
  - [x] on the new registration page, add button to either authorize or create a registration link
  - [x] create a page that recognises token and secret parameters that forms a welcome page for a subject
  - [x] create a page that thanks the subject for finishing registration or warn them of errors
  - [x] add registration action to each row of a user that is unauthorised, so unauthorised users can be re-authorized.

The API can be used as follows:

- press create user
- select source type, project, user ID, start date, end date
- request `POST <backend url>/users` with contents
    ```ts
    interface RestSourceUserRequest {
        projectId: String
        userId: String
        startDate: Instant
        endDate?: Instant
        sourceType: String
    }
    ```
  which will have response (for user 1)
    ```
    HTTP 201 Created
    Location: <backend url>/users/1
    ```
    ```ts
    interface RestSourceUserResponse {
        id: String
        projectId: String
        userId: String
        sourceId: String
        startDate: Instant
        endDate?: Instant
        sourceType: String
        createdAt: Instant
        humanReadableUserId: String
        externalId: String
        serviceUserId?: String
        isAuthorized: Boolean
        timesReset: Int
        version: String
    }
    ```

- After the user has been created, an account registration request can be created at `POST <backend url>/registrations` with contents

     ```ts
     interface RegistrationCreateRequest {
         userId: String  // matches RestSourceUserResponse.id
         persistent?: Boolean  // set to true to get a long-living token
     }
     ```

  with response (for token `abcdef`)

    ```
    HTTP 201 Created
    Location: <backend url>/registrations/abcdef
    ```
     ```ts
     interface RegistrationResponse {
         token: String
         secret?: String  // only defined if the registration is persistent
         userId: String
         authEndpointUrl?: String  // only defined if the registration is not persistent
         expiresAt: Instant
         persistent: Boolean
     }
     ```

- The token and secret can be used in a frontend URL to pass to the backend at a later time, or the `authEndpointUrl` can used immediately.

    - If the user created a persistent token, the `authEndpointUrl` can be retrieved with a request to `POST <backend url>/registrations/<token>` with contents

        ```ts
        interface RegistrationRequest {
            secret: String
        }
        ```

      and it will return the authorization URL to be used:

        ```ts
        interface RegistrationResponse {
            token: String
            userId: String
            authEndpointUrl: String
            expiresAt: Instant
            persistent: Boolean
        }
        ```

- When the user is authorized, they will be redirected to `<frontend url>/users:new` with a number of query parameters, including the `state` parameter. To finalize the authentication procedure, call `POST <backend url>/registrations/<state>/authorize` with contents. Alternatively, if the external service does not provide the state back, store the token in session storage and use that instead.

    ```ts
    interface AuthorizeRequest {
        code?: String
        oauth_token?: String
        oauth_verifier?: String
        oauth_token_secret?: String
    }
    ```
    with response

    ```
    HTTP 201 Created
    Location: <backend url>/source-clients/<sourceType>/authorization/<serviceUserId>
    ```
    ```ts
    interface RegistrationResponse {
         token: String
         userId: String
         expiresAt: Instant
         persistent: Boolean
    }
    ```

    - If the response indicated that the token was persistent, the user that can simply be thanked for entering their user details.
    - Otherwise, the user can be redirected to the list of users.
    - In both cases, the registration is now removed.


authorizer-app-backend/src/main/java/org/radarbase/authorizer/Config.kt
```diff
@@ -32,13 +35,39 @@ data class Config(
 
 data class AuthorizerServiceConfig(
     var baseUri: URI = URI.create("http://0.0.0.0:8080/rest-sources/backend/"),
-    var advertisedBaseUri: URI? = null,
+    val advertisedBaseUri: URI? = null,
+    val frontendBaseUri: URI? = null,
```
nivemaham, line 39: what is the difference between the two? one is backend and the other is frontend? (both are advertised ones I assume?)

authorizer-app-backend/src/main/java/org/radarbase/authorizer/doa/RestSourceUserRepositoryImpl.kt
```diff
@@ -23,61 +23,62 @@ import org.radarbase.authorizer.api.Page
...
-        val existingUser = createQuery(queryString, RestSourceUser::class.java)
-            .setParameter("sourceType", sourceType)
-            .setParameter("externalUserId", externalUserId)
+    override fun create(user: RestSourceUserDTO): RestSourceUser = transact {
+        val existingUser = createQuery(
+            """
+            SELECT u
+            FROM RestSourceUser u
+            WHERE u.sourceType = :sourceType
+                AND u.userId = :userId
```
nivemaham, line 45: Isn't the idea to look for existing users with the same fitbit account (externalId) ? or did we rename the externalUserId to userId?

authorizer-app-backend/src/main/java/org/radarbase/authorizer/Config.kt
blootsvoets: Indeed, the first one is the advertisedBaseUri for the current service; the other is where the frontend can be contacted by clients.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/doa/RestSourceUserRepositoryImpl.kt
blootsvoets: In this altered flow, the external ID is not yet known when we create the user. It is set after a new rest source user (combination MP user + source type) is made.

PR #153, opened 2021-09-06
Update OAuth2

blootsvoets:
Adds `redirect_uri` whenever requesting an OAuth 2.0 access token. This is now required by e.g. the Fitbit API.


nivemaham (commented):
i can't see the OAuth2 part

blootsvoets:
It's mainly in OAuth2RestSourceAuthorizationService.kt, but also in Config.kt

authorizer-app-backend/build.gradle.kts
```diff
@@ -11,6 +12,10 @@ plugins {
 
 application {
     mainClass.set("org.radarbase.authorizer.Main")
+    applicationDefaultJvmArgs = listOf(
+        "-Djava.security.egd=file:/dev/./urandom",
```
nivemaham, line 16: Why do we use this? Just for my interest.

nivemaham (approved):
LGTM

authorizer-app-backend/build.gradle.kts
blootsvoets: It tells Java that it shouldn't rely on the entropy of the machine (on VM's there is typically too little entropy), but use a pseudorandom generator. This avoids any delays in SSL processing where randomness is required -- see e.g. https://www.baeldung.com/java-security-egd#10-what-about-virtual-machines

PR #159, opened 2021-09-09
Lock refresh token

blootsvoets:
Ensure that refresh tokens are not requested simultaneously. This is especially important because if an access token is requested twice with the same refresh token, the second call returns that the refresh token is no longer valid. This will invalidate the entire authorization.


authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RestSourceUserResource.kt
```diff
@@ -191,23 +194,27 @@ class RestSourceUserResource(
...
     }
 
     @POST
     @Path("{id}/token")
     @NeedsPermission(Permission.Entity.MEASUREMENT, Permission.Operation.CREATE)
     fun refreshToken(@PathParam("id") userId: Long): TokenDTO {
-        val user = ensureUser(userId)
-        auth.checkPermissionOnSubject(Permission.MEASUREMENT_CREATE, user.projectId, user.userId)
-        return refreshToken(userId, user)
+        return lockService.runLocked("token-$userId", Duration.ofSeconds(10)) {
```
yatharthranjan, line 213: Would this be better placed inside the `refreshToken(userId: Long, user: RestSourceUser)` function. So whenever it is called, it is always locked. And then it can be placed around the refresh token part of that function (so the client does not have to wait for a lock if there is an exception to be thrown).

yatharthranjan (commented):
Looks fine to me, just a minor comment.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RestSourceUserResource.kt
blootsvoets: The issue is that the subjects may be updated after the lock is achieved, so ensureUser always gets the latest user, including that latest tokens, etc. Indeed I could do a second user fetch.

PR #193, opened 2022-09-06
Fix OAuth signature encoding

mpgxvii:
- The encoded params for the signature needs the null values to be 'null' otherwise the access token request returns a `401` error
- This changes the null param values to be null instead of a blank string


blootsvoets (commented):
LGTM

PR #194, opened 2022-09-07
Fix retrieving and storing of OAuth1 auth params

mpgxvii:
- The OAuth1 authentication needs the params: `oauth_token`, `oauth_token_secret`, and `oauth_token_verifier`, which is retrieved at different steps of the authentication process. These need to be stored so that they can be used at the final step which is requesting the access token. 
- Specifically, the `oauth_token_secret` was null (and this is needed as part of the signing key) so the signature was invalid.

### DIFF
--- a/authorizer-app/src/app/admin/services/user.service.ts
+++ b/authorizer-app/src/app/admin/services/user.service.ts
@@ -10,13 +10,12 @@ import {
 } from '@app/admin/models/rest-source-user.model';
 
 import {environment} from '@environments/environment';
+import { StorageItem } from '@app/shared/enums/storage-item';
 
 @Injectable({
   providedIn: 'root'
 })
 export class UserService {
-  AUTH_ENDPOINT_PARAMS_STORAGE_KEY = 'auth_endpoint_params';
-
   constructor(private http: HttpClient) {}
 
   getUsersOfProject(projectId: string): Observable<RestSourceUsers> {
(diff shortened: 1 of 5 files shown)


authorizer-app/src/app/admin/services/user.service.ts
```diff
@@ -68,4 +69,28 @@ export class UserService {
     );
     return this.http.delete(url);
   }
+
+  storeUserAuthParams(url: string) {
+    const params = this.getJsonFromUrl(url);
+    localStorage.setItem(
+      this.AUTH_ENDPOINT_PARAMS_STORAGE_KEY,
```
yatharthranjan, line 76: Maybe add the key to enums in `StorageItem` in `"@app/shared/enums/storage-item"` so it can be accessed from outside the class.

authorizer-app/src/app/shared/containers/authorization-complete-page/authorization-complete-page.component.ts
```diff
@@ -27,7 +27,22 @@ export class AuthorizationCompletePageComponent implements OnInit {
...
+      storedParams.oauth_token
+    );
+    const oauth_verifier = this.getOrDefault(
+      queryParams.oauth_verifier,
+      storedParams.oauth_verifier
+    );
+    const oauth_token_secret = this.getOrDefault(
+      queryParams.oauth_token_secret,
+      storedParams.oauth_token_secret
+    );
```
yatharthranjan, line 44: Should we delete these items from storage after the authorization is complete to avoid it interfering with future registrations?

authorizer-app/src/app/admin/services/user.service.ts
```diff
@@ -68,4 +69,28 @@ export class UserService {
     );
     return this.http.delete(url);
   }
+
+  storeUserAuthParams(url: string) {
```
yatharthranjan, line 73: I see this used to be present in 3.x.x versions -- https://github.com/RADAR-base/RADAR-Rest-Source-Auth/blob/97fd30092b4e52a1f5ad7e26783f1d76cd6941db/authorizer-app/src/app/services/source-client-authorization.service.ts#L48

I wonder why it was removed during refactoring.

yatharthranjan (commented):
Thanks for investigating the issue. Please find some comments.

yatharthranjan (approved):
LGTM, thanks for the changes

PR #214, opened 2022-10-27
only return authorized users by default

### LINKED ISSUE #211: Make authorised query param true by default when GET users
This is to avoid breaking functionality in existing apps (push-endpoint).

Here is the line where change is required
https://github.com/RADAR-base/RADAR-Rest-Source-Auth/blob/28b04d49e0882dd2dac9fdef6d2ed72d13dad998/authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RestSourceUserResource.kt#L60

yatharthranjan:
- Closes #211


mpgxvii (approved):
LGTM

blootsvoets:
This is the PR that caused #225. It should be fixed in the latest release.

PR #216, opened 2022-10-28
Release 4.1.1

yatharthranjan:
- Only return authorised users by default on the `GET /users` endpoint
- Remove authentication from deregister endpoint (only garmin supported)


mpgxvii (approved):
LGTM

Issue #225, opened 2023-01-27
User exists error but UI shows user as ready for authorizing


yatharthranjan:
We have a user who is shown as ok to authorize in the UI, but once we start Authorization, the following exception is thrown - 

![SCR-20230127-h72](https://user-images.githubusercontent.com/11093544/215086216-825859a1-1b71-4bff-8220-0fb56be24fdb.png)

![image](https://user-images.githubusercontent.com/11093544/215086210-4f8a07ea-0a53-4534-a6e0-14e592a5a3ad.png)

Logs - 

```
[2023-01-26 20:25:30,469] ERROR - Rolling back operation (HibernateRepository.kt:54)
org.radarbase.jersey.exception.HttpConflictException: [409] user_exists: User 5dffc387-7a1b-430f-bad4-a61c384d4821 already exists.
	at org.radarbase.authorizer.doa.RestSourceUserRepositoryImpl$create$1.invoke(RestSourceUserRepositoryImpl.kt:53) ~[authorizer-app-backend-4.1.1.jar:?]
	at org.radarbase.authorizer.doa.RestSourceUserRepositoryImpl$create$1.invoke(RestSourceUserRepositoryImpl.kt:38) ~[authorizer-app-backend-4.1.1.jar:?]
	at org.radarbase.jersey.hibernate.HibernateRepository$transact$1.invoke(HibernateRepository.kt:26) ~[radar-jersey-hibernate-0.9.0.jar:0.9.0]
	at org.radarbase.jersey.hibernate.HibernateRepository$transact$1.invoke(HibernateRepository.kt:25) ~[radar-jersey-hibernate-0.9.0.jar:0.9.0]
	at org.radarbase.jersey.hibernate.HibernateRepository.createTransaction(HibernateRepository.kt:38) ~[radar-jersey-hibernate-0.9.0.jar:0.9.0]
	at org.radarbase.jersey.hibernate.HibernateRepository.transact(HibernateRepository.kt:25) ~[radar-jersey-hibernate-0.9.0.jar:0.9.0]
	at org.radarbase.jersey.hibernate.HibernateRepository.transact(HibernateRepository.kt:18) ~[radar-jersey-hibernate-0.9.0.jar:0.9.0]
	at org.radarbase.authorizer.doa.RestSourceUserRepositoryImpl.create(RestSourceUserRepositoryImpl.kt:38) ~[authorizer-app-backend-4.1.1.jar:?]
[...]

PR #228, opened 2023-03-01
Ignore checksum check for one change

yatharthranjan:
Currently, a change in checksum is preventing upgrade of the app. Based on discussion in #225


blootsvoets (approved):
Thanks for the update

Issue #234, opened 2023-05-09
Authorization using Generate URL not working for Garmin


yatharthranjan:
Authorization using the link generated from Generate URL button for Garmin source-type does not work. This works for Fitbit source type.

Not much info in the logs- 

```
ERROR - [407] GET registrations/08APtrwC8WZev_hW7wddrq4h/authorize - user_unauthorized: Access token can not be retrieved (HttpApplicationExceptionMapper.kt:30)
```

<img width="1261" alt="image" src="https://github.com/RADAR-base/RADAR-Rest-Source-Auth/assets/11093544/0ef3eacc-eaa6-439c-8dbb-2c97d617aacd">

here is the url format - 

```
https://trials.radar-base.net/rest-sources/authorizer/users:new?state=KyWl94iZjNL55zem4mSizbcr&oauth_token=84xxxxxxx-79ee-xxxxxxxx-91bxxxxxxxx4ae5&oauth_verifier=MiVxxxx3C
```

blootsvoets:
Fixed by #235

PR #235, opened 2023-05-10
Store auth params in authorization page 

mpgxvii:
- Authentication with OAuth1/Garmin fails because the token secret doesn't get stored when an auth URL is generated
- This fixes it by storing the token secret on the authorization page side

Solves #234


yatharthranjan (approved):
LGTM

PR #241, opened 2023-10-12
Oura Oauth2 Integration

Hsankesara:
Extended OAuth2 and created new Oura Oauth2 integration.

### DIFF
--- a/authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OuraAuthorizationService.kt
+++ b/authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OuraAuthorizationService.kt
@@ -7,6 +7,7 @@ import okhttp3.Credentials
 import okhttp3.FormBody
 import okhttp3.OkHttpClient
 import okhttp3.Request
+import okhttp3.Response
 import org.radarbase.authorizer.api.RequestTokenPayload
 import org.radarbase.authorizer.api.RestOauth2AccessToken
 import org.radarbase.authorizer.api.SignRequestParams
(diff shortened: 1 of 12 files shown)


authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OuraAuthorizationService.kt
```diff
@@ -0,0 +1,50 @@
...
+        val authConfig = clients.forSourceType(user.sourceType)
+        val deregistrationEndpoint = checkNotNull(authConfig.deregistrationEndpoint)
+
+        val revokeURI = UriBuilder.fromUri(deregistrationEndpoint).queryParam("access_token", accessToken).build().toString()
+
+        val credentials = Credentials.basic(
+            checkNotNull(authConfig.clientId),
+            checkNotNull(authConfig.clientSecret),
+        )
+        val form = FormBody.Builder().add("token", accessToken).build()
```
mpgxvii, line 36: I don't think we need this anymore since the `accessToken` will be passed via query params.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OuraAuthorizationService.kt
```diff
@@ -0,0 +1,50 @@
...
+            url(revokeURI)
+        }.post(form).build()
+        val response = httpClient.request(requestObj)
+        if (response) {
+            logger.info("Successfully revoked token for user {}", user.userId)
+        } else {
+            logger.error("Failed to revoke token for user {}", user.userId)
+            return false
+        }
+        return httpClient.request(post(form, user.sourceType))
```
mpgxvii, line 48: I think the request is already being made in line 41? So no need for this? But we would still need to return a `boolean`.

docker/etc/rest-source-authorizer/authorizer.yml
```diff
@@ -19,10 +19,14 @@ database:
...
+  uri: redis://redis:6379
+
 restSourceClients:
-  - sourceType: FitBit
-    authorizationEndpoint: https://www.fitbit.com/oauth2/authorize
-    tokenEndpoint: https://api.fitbit.com/oauth2/token
-    clientId: Fitbit-clientid
-    clientSecret: Fitbit-clientsecret
-    scope: activity heartrate sleep profile
+  - sourceType: Oura
```
mpgxvii, line 26: I think we can just keep `Fitbit` here as well? Just like in the `authorizer.yml.template`.

mpgxvii (changes requested):
Looks good, some minor comments/questions.

Hsankesara:
Thanks, @mpgxvii, I have made the changes.

docker/etc/rest-source-authorizer/authorizer.yml
mpgxvii: Sorry @Hsankesara, I meant can we keep both Fitbit and Oura here?

Hsankesara: Ah sorry, my bad. That makes more sense. Will update it

docker/etc/rest-source-authorizer/authorizer.yml.template
```diff
@@ -41,4 +41,10 @@ restSourceClients:
...
     clientSecret: Garmin-clientsecret
     scope: activity heartrate sleep profile
-
+  - sourceType: Oura
+    authorizationEndpoint: https://cloud.ouraring.com/oauth/authorize
+    tokenEndpoint: https://api.ouraring.com/oauth/token
+    deregistrationEndpoint: https://api.ouraring.com/oauth/revoke
+    clientId: Oura-clientid
+    clientSecret: Oura-clinetsecret
+    scope: daily session heartrate workout tag personal email spo2
```
mpgxvii, line 50: Can we also include the `ring_configuration` scope?

Hsankesara: Oh didn't know about this scope. Added it now.

mpgxvii: Thanks!

mpgxvii:
@Hsankesara do we have a `serviceUserId` or equivalent that we receive from Oura that identifies the users? Similar to Fitbit?

Hsankesara:
> @Hsankesara do we have a `serviceUserId` or equivalent that we receive from Oura that identifies the users? Similar to Fitbit?

I don't think so. it doesn't seem to returning any service_id equivalent.

mpgxvii (approved):
LGTM, thanks for the changes!

Hsankesara:
Thanks, @mpgxvii!

yatharthranjan:
> I don't think so. it doesn't seem to returning any service_id equivalent.

We need this in the push-endpoint as the webhook only provides a user_id for mapping to users in our user repository. Seems it is on the API path https://api.ouraring.com/v1/userinfo?access_token=xyz , we should get this and add it to user.

PR #243, opened 2023-10-18
Fetching Oura service user id

Hsankesara:
Added function to add Oura service user ID while calling `requestAccessToken`


authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OuraAuthorizationService.kt
```diff
@@ -51,4 +77,7 @@ class OuraAuthorizationService(
             return false
         }
     }
+    companion object {
+        private const val Oura_USER_ID_ENDPOINT = "https://api.ouraring.com/v1/userinfo?"
```
mpgxvii, line 81: Should this be renamed to `OURA_USER_ID_ENDPOINT`?

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OuraAuthorizationService.kt
```diff
@@ -3,20 +3,54 @@ package org.radarbase.authorizer.service
...
+            add("grant_type", "authorization_code")
+            add("client_id", clientId)
+            add("redirect_uri", config.service.callbackUrl.toString())
+        }.build()
+        val accessToken: RestOauth2AccessToken = httpClient.requestJson(post(form, sourceType), tokenReader)
+        if (accessToken.accessToken == null) {
+            logger.error("Failed to get access token for user {}", clientId)
+            throw HttpBadGatewayException("Service $sourceType did not provide a result")
+        }
+        val ouraUserUri = UriBuilder.fromUri(Oura_USER_ID_ENDPOINT).queryParam("access_token", accessToken.accessToken).build().toString()
```
mpgxvii, line 42: Could we move this to a separate function and have the `accessToken` as the parameter?

mpgxvii (changes requested):

Hsankesara:
Thanks @mpgxvii, i've made the code changes. Have a look.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OuraAuthorizationService.kt
```diff
@@ -51,4 +64,17 @@ class OuraAuthorizationService(
             return false
         }
     }
+
+    private fun getExternalId(accessToken: String): String {
+        val ouraUserUri = UriBuilder.fromUri(OURA_USER_ID_ENDPOINT).queryParam("access_token", accessToken).build().toString()
+        val userReq = Request.Builder().apply {
+            url(ouraUserUri)
+        }.build()
+        val userIdObj: OuraAuthUserId = httpClient.requestJson(userReq, oauthUserReader)
```
mpgxvii, line 73: I think we can also add additional response/error handling here in case the request is unsuccessful.

Hsankesara: not sure how to do error handling after calling `requestJson`. Other option would be to execute request using this `client.newCall(request).execute()` and handle error like in `GarminSourceAuthorizationService.kt`. Happy to do it unless there is a better way to do the same thing

mpgxvii: Yes can we just switch to the implementation similar to `GarminSourceAuthorizationService`. Because in case in the future when the API gets deprecated or there are issues with this endpoint, we can handle the error. (Since this is using v1 of the API)

mpgxvii (approved):
LGTM, thanks @Hsankesara @blootsvoets for the updates!

PR #244, opened 2023-10-24
Update radar-auth

blootsvoets:
- Use coroutines fetch
- Use radar-auth 2.1.0
- Use radar-commons-kotlin 1.1.1
- Use ktor with kotlinx-serialization
- Use `buildSrc` and `radar-commons-gradle` for building
- Use async for serving responses. Especially relevant for instances with a lot of users in the connector.


authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OuraAuthorizationService.kt
```diff
@@ -27,28 +23,34 @@ class OuraAuthorizationService(
...
-
-        var requestObj = Request.Builder().apply {
-            url(revokeURI)
-            post(FormBody.Builder().build())
-            header("Authorization", credentials)
-            header("Content-Type", "application/x-www-form-urlencoded")
-            header("Accept", "application/json")
-        }.build()
+        val isSuccess = try {
+            val response = httpClient.submitForm {
```
mpgxvii, line 27: I haven't really used `withContext(Dispatchers.IO)` much, but should it be used here as well?

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OAuth1RestSourceAuthorizationService.kt
```diff
@@ -79,111 +94,112 @@ abstract class OAuth1RestSourceAuthorizationService(
...
         params[OAUTH_SIGNATURE] =
             OauthSignature(url, params, method, authConfig.clientSecret, tokens.tokenSecret).getEncodedSignature()
-        val headers = params.toFormattedHeader()
 
-        return Request.Builder()
-            .url(url)
-            .header("Authorization", "OAuth $headers")
-            .method(method, if (method == "POST") "".toRequestBody(null) else null)
-            .build()
+        return httpClient.request(url = Url(url)) {
```
mpgxvii, line 194: If we run `withContext` here, and remove it from the methods that call `request()`, will those still run in the context? Or would we need to still run `withContext` in the outer methods?

mpgxvii (dismissed):
LGTM, thanks!

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OuraAuthorizationService.kt
blootsvoets: Good find, yes that's right.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OAuth1RestSourceAuthorizationService.kt
blootsvoets: yes, you're right, I've added it to the lowermost function now. Indeed `withContext` calls stack without problem. If it's already in the right context it won't do any switches.

PR #245, opened 2023-10-25
Release 4.4.0

mpgxvii:
- Stores auth params in authorization page temporarily to be used later on in auth process (#235)
- Updates radar-auth, updates to use async for serving responses, and dependency updates (#244)
- Adds Oura Oauth2 Integration (#241, #243)


blootsvoets (dismissed):
LGTM. @peyman-mohtashami @Bdegraaf1234 would you take a look as well?

yatharthranjan (approved):
LGTM, will test further on stage env.

PR #247, opened 2023-10-25
solved error of fetching external id in Oura

Hsankesara:


authorizer-app-backend/src/main/java/org/radarbase/authorizer/api/ApiDeclarations.kt
```diff
@@ -52,9 +52,9 @@ data class RestOauth1UserId(
 
 @Serializable
 data class OuraAuthUserId(
-    val age: String,
-    val weight: String,
-    val height: String,
+    val age: Int,
+    val weight: Float,
+    val height: Int,
```
mpgxvii, line 57: Can this be `float` as well?

mpgxvii (approved):
LGTM

authorizer-app-backend/src/main/java/org/radarbase/authorizer/api/ApiDeclarations.kt
Hsankesara: can do that, but I think Oura returns height in the metric unit (in integer format). But making it float shouldn't change anything as we are not collecting this data

mpgxvii: Ah ok I wasn't sure if the height would return decimals as well. But yes we can keep it Float.

PR #263, opened 2024-02-28
pass client_id in token_request

Bdegraaf1234:
adapt the token request in the front end to accomodate ory

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -126,9 +126,9 @@ jobs:
           # Use runtime labels from docker_meta_backend as well as fixed labels
           labels: |
             ${{ steps.docker_meta.outputs.labels }}
-            maintainer=Joris Borgdorff <joris@thehyve.nl>, Nivethika Mahasivam <nivethika@thehyve.nl>, Pauline Conde <pauline.conde@kcl.ac.uk>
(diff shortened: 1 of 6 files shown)


peyman-mohtashami (approved):

Bdegraaf1234:
@mpgxvii 

hydra does want or at least allows the client_id in the post request
https://www.ory.sh/docs/hydra/reference/api#tag/oAuth2/operation/oauth2TokenExchange

The complete technical spec for oauth can be found [here](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3), I linked the section on the authorization code grant, it looks like if you authenticate with basic you don't have to provide the client ID in the post request but you can.


[This](https://github.com/RADAR-base/ManagementPortal/commit/128b960061f884bbd15b41a7bdfb784a6754efdc#diff-ac7c296fb7985e2eb39d9f5bbb311f4f443707360f5e6291a77c279fdcd81131) is the commit where I messed with the token endpoint, it is a bit hacky so I am happy it will be removed, but I think the implementation on the RSA side stands :)

mpgxvii:
@Bdegraaf1234 Oh okay I see. Thanks for the clarification! :)

PR #274, opened 2024-07-05
Bump radar-jersey to SNAPSHOT version

Bdegraaf1234:
Fix the problems with radar-jersey returning malformed `www-authenticate` headers.

See:
https://github.com/RADAR-base/radar-jersey/pull/88


pvannierop (dismissed):
Please remove the mavenLocal() repo.

PR #282, opened 2024-09-26
Add support for Ory auth

mpgxvii:
- Adds support for Ory auth in front-end
- When creating a `RestSourceUser` and it already exists, return the existing user (needed when for external authorisation e.g. SEP user)
- Create custom `ManagementPortalResourceEnhancer` to bind custom MPClient
- Also created custom `MPClientFactory` to update client credentials config (to work with Ory hydra)
- Depends on https://github.com/RADAR-base/radar-commons/pull/176


authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/MPClient.kt
```diff
@@ -0,0 +1,153 @@
...
+        HttpClient(CIO) {
+            install(ContentNegotiation) {
+                json(Json { ignoreUnknownKeys = true })
+            }
+            install(HttpTimeout) {
+                requestTimeoutMillis = 30_000
+            }
+        }
+
+    private var accessToken: String? = null
```
yatharthranjan, line 45: use lateinit

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -6,7 +6,7 @@ object Versions {
 
     const val kotlin = "1.9.23"
 
-    const val radarCommons = "1.1.2"
+    const val radarCommons = "1.1.3-SNAPSHOT"
```
mpgxvii, line 9: Will update when released.

yatharthranjan (approved):
Looks great -- reusing the jersey code. Thanks!

yatharthranjan (approved):
LGTM

PR #292, opened 2025-01-16
Add support for additional params in auth url

mpgxvii:
- Add support for additional params in the `AuthorizationPage` to allow redirecting directly to Fitbit auth and allow redirecting to a custom URL after authorization is complete (using `redirect=true` and `redirect_uri=URL-HERE`
- If there is an error, it will redirect with `?error=message` param
- Also added improvements to the `AuthorzationCompletePage`
- Needed by the SEP portal (since user will self authorise and need to be redirected back to the SEP portal)

### DIFF
--- a/authorizer-app/src/app/admin/services/user.service.ts
+++ b/authorizer-app/src/app/admin/services/user.service.ts
@@ -70,6 +70,26 @@ export class UserService {
     return this.http.delete(url);
   }
 
+  storeAuthorizationToken(token: string) {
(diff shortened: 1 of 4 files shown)


yatharthranjan:
Just reported by callum, if there is an error when linking fitbit (eg - user has already connected fitbit ), then the error is still showin in rest-sources-auth UI. Ideally, it should redirect directly back to the SEP UI if redirect=true is set in the params, and should include the error in the url params.

Issue #279, opened 2024-08-29
Inconsistent state with rest source and MP users


mpgxvii:
- When a `ManagementPortal` user/subject is deleted, this is not deleted on the `rest-sources-auth` side which results in an inconsistent state
- So the Kafka connectors still end up trying to get the tokens for the rest source users, resulting in `HttpException` in the connector

yatharthranjan:
I think this might be due to caching on the two systems -

1. Cached on the rest-sources side
2. Cached on the user repository imp in the connectors

I would try reducing the cache expiry and testing. Additionally, I would have thought this would be a transient issue, i.e., it should resolve itself once the cache has been updated (unless the source task is not reconfiguring on update of users).

yatharthranjan:
Yes, a periodic clean up task on rest-sources auth would be good to sync state with MP.

Should the appserver do this too? Otherwise it might keep sending notifications to users that are deleted from the MP but still have the app installed?

Hsankesara:
Think we faced the same issue yesterday. I think I have some bandwidth, happy to work on it. Would it makes sense that when it throw this particular error, it can run a cleanup and restart the task?

Hsankesara:
> We had this problem on CONNECT project since a couple of days. It broke Fitbit data collection.

In the end, we have to delete the participant from the DB.

mpgxvii:
Solved and released in https://github.com/RADAR-base/RADAR-Rest-Source-Auth/pull/314

pvannierop:
I think that we have deployed this fix, and the error still occurs after this. Moreover, we should think about the situation where a user is offboarded, but not all Fitbit data collected yet?

PR #312, opened 2025-09-26
Add periodic user cleanup

mpgxvii:
- Add UserSyncLifecycleManager to periodically cleanup rest source users that have been deleted in Management Portal to ensure consistent state

Solves #279


authorizer-app-backend/src/main/java/org/radarbase/authorizer/lifecycle/UserSyncLifecycleManager.kt
```diff
@@ -0,0 +1,115 @@
...
+
+@Provider
+@Singleton
+class UserSyncLifecycleManager(
+    @BackgroundScheduler
+    @Context
+    private val scheduler: ScheduledExecutorService,
+    @Context private val userRepository: RestSourceUserRepository,
+    @Context private val projectService: RadarProjectService,
+    @Context private val authorizationService: RestSourceAuthorizationService,
```
this-Aditya, line 30: Hi @mpgxvii, the `authorizationService` is unused and can be removed. Aside from that, the PR looks good to me.

mpgxvii: Removed now 👍

PR #306, opened 2025-07-22
Fix duplicate service ID

yatharthranjan:
Attempts to fix #305

PR #314, opened 2025-10-03
Release 4.4.12

mpgxvii:
- Add periodic user cleanup to cleanup rest source users that have been deleted in Management Portal to ensure consistent state (#312)
- Fix duplicate service ID issue so that a device/source account can only be assigned to a single user (#306)


this-Aditya (approved):
LGTM

PR #320, opened 2026-03-12
Release 4.4.14 (March 2026 security updates)

pvannierop:
This PR will:
- Impl. Gradle version catalog
- Up Gradle to 8.14
- Apply March 2026 security updates
- Add support for Sentry log monitoring


build.gradle.kts
```diff
@@ -1,28 +1,25 @@
...
 
 subprojects {
     apply(plugin = "org.radarbase.radar-kotlin")
 
     radarKotlin {
-        javaVersion.set(Versions.java)
-        kotlinVersion.set(Versions.kotlin)
-        slf4jVersion.set(Versions.slf4j)
-        log4j2Version.set(Versions.log4j2)
-        junitVersion.set(Versions.junit)
```
this-Aditya, line 26: If we skip setting the versions here for Java, Kotlin, etc., then these versions will be controlled entirely by the radar-kotlin plugin defaults. Do we need this behavior?

pvannierop: @this-Aditya  I removed all these statement, but added the sentry related ones.

build.gradle.kts
```diff
@@ -20,5 +20,7 @@ subprojects {
 
     radarKotlin {
         log4j2Version.set(rootProject.libs.versions.log4j2)
+        sentryEnabled.set(true)
+        openTelemetryAgentEnabled.set(false)
```
this-Aditya, line 24: It is false by default, should we remove this? Or do you think it is better to be explicit?

this-Aditya (commented):
Thanks @pvannierop, one last question:

build.gradle.kts
pvannierop: It is false by default, but would like to show everywhere that we can enable it at some point. I have added it throughout he code base.

this-Aditya: Okay, thanks for the clarification.

this-Aditya (approved):

PR #321, opened 2026-03-23
Fix missing user_id during token refresh

mpgxvii:
- Fallback to existing userId if not present when refreshing token


Hsankesara (approved):
LGTM! 👍

PR #322, opened 2026-03-25
Add project ID to account already registered error message

this-Aditya:
Added project ID to the error message when a user tries to register a device account that is already linked to another user, making it easier to identify  which project the existing registration belongs to.


mpgxvii (approved):
LGTM! Thanks, it would indeed be helpful to know which project the user belongs to.

PR #317, opened 2026-02-04
Garmin OAuth migration from OAuth 1.0 to OAuth 2.0

this-Aditya:
Fixes #313

This PR migrates Garmin authorization from OAuth 1.0 to OAuth 2.0. Note that this change must be coordinated with the radar push endpoint, as the data structure for both push and ping payloads will change after the migration.

PR #324, opened 2026-03-26
Release 4.4.15

mpgxvii:
- Add project ID to account already registered error message (#322)
- Fix missing Oura user_id during token refresh (#321)
- Add Garmin Oauth2 Authorization (#317)


this-Aditya (dismissed):

this-Aditya:
One point I think is worth mentioning: the master branch is ahead of dev, and master publishes docker images to GHCR, while dev publishes them to dockerhub. That’s why we’re not able to see any dev images in packages. Should we also sync master and dev before/after this release so that images are pushed to GHCR?

this-Aditya:
@copilot apply changes based on the comments in [this thread](https://github.com/RADAR-base/RADAR-Rest-Source-Auth/pull/324#pullrequestreview-4021930340)

this-Aditya (dismissed):

authorizer-app-backend/src/main/java/org/radarbase/authorizer/config/RestSourceClient.kt
```diff
@@ -14,10 +14,12 @@ data class RestSourceClient(
     val grantType: String? = null,
     val scope: String? = null,
     val state: String? = null,
+    val usesPkce: Boolean = false,
+    val oauthVersion: String = "oauth2",
```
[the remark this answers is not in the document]
this-Aditya: No, oauth2 should be default.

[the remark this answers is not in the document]
yatharthranjan: perhaps something to note in release notes, like

>BREAKING CHANGES, Garmin defaults to Oauth2.0
if still using oauth1.0 configure `oauthVersion = "oauth1"`

authorizer-app-backend/src/main/java/org/radarbase/authorizer/enhancer/AuthorizerResourceEnhancer.kt
```diff
@@ -102,10 +112,18 @@ class AuthorizerResourceEnhancer(
         bind(DelegatedRestSourceAuthorizationService::class.java)
             .to(RestSourceAuthorizationService::class.java)
 
-        bind(GarminSourceAuthorizationService::class.java)
-            .to(RestSourceAuthorizationService::class.java)
-            .named(GARMIN_AUTH)
-            .`in`(Singleton::class.java)
+        // Bind Garmin service based on a configured oauthVersion: "oauth2" → PKCE flow, "oauth1" → legacy flow.
+        if (sourceTypeOauthMap[GARMIN_AUTH].equals("oauth2", ignoreCase = true)) {
```
yatharthranjan: do you really need a map here? why not directly check using the config key?

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/GarminOAuth2AuthorizationService.kt
```diff
@@ -0,0 +1,264 @@
...
+            parameters.append("redirect_uri", config.service.callbackUrl.toString())
+            buildString()
+        }
+    }
+
+    private fun checkForUsersWithElapsedEndDateAndDeregister() {
+        asyncService.runBlocking {
+            userRepository
+                .queryAllWithElapsedEndDate(GARMIN_AUTH)
+                .forkJoin { revokeToken(it) }
```
yatharthranjan, line 230: just discussed, should check if access token exists here before revoking it

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/GarminOAuth2AuthorizationService.kt
```diff
@@ -0,0 +1,264 @@
...
+        }
+    }
+
+    override suspend fun deregisterUser(user: RestSourceUser) {
+        userRepository.delete(user)
```
yatharthranjan, line 235: do we need to deregister on Garmin too (call the deleteRegistration function) ?

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/GarminOAuth2AuthorizationService.kt
```diff
@@ -0,0 +1,264 @@
...
+            headers {
+                append(HttpHeaders.Authorization, "Bearer $accessToken")
+            }
+        }
+
+        when (response.status) {
+            HttpStatusCode.OK -> response.body<RestOauth2UserId>().userId
+
+            HttpStatusCode.BadRequest, HttpStatusCode.Unauthorized, HttpStatusCode.Forbidden -> throw HttpBadGatewayException(
+                "Service was unable to fetch the external ID",
```
yatharthranjan: maybe good to include original exception stack?

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/RestSourceAuthorizationService.kt
```diff
@@ -31,7 +31,7 @@ import org.radarbase.authorizer.doa.entity.RestSourceUser
 import kotlin.time.Duration.Companion.seconds
 
 interface RestSourceAuthorizationService {
-    suspend fun requestAccessToken(payload: RequestTokenPayload, sourceType: String): RestOauth2AccessToken
+    suspend fun requestAccessToken(payload: RequestTokenPayload, sourceType: String, token: String? = null): RestOauth2AccessToken
```
yatharthranjan: Probably better to use method overloading and add a new function with the token, this way you wouldn't need to modify existing code.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OAuth1RestSourceAuthorizationService.kt
```diff
@@ -58,7 +58,7 @@ abstract class OAuth1RestSourceAuthorizationService(
 ) : RestSourceAuthorizationService {
     private val httpClient = RestSourceAuthorizationService.httpClient()
 
-    override suspend fun requestAccessToken(payload: RequestTokenPayload, sourceType: String): RestOauth2AccessToken {
+    override suspend fun requestAccessToken(payload: RequestTokenPayload, sourceType: String, token: String?): RestOauth2AccessToken {
```
yatharthranjan: can be an overloaded function and here you can just

```kt
override suspend fun requestAccessToken(payload: RequestTokenPayload, sourceType: String): RestOauth2AccessToken {
.....   // keep as is
}

override suspend fun requestAccessToken(payload: RequestTokenPayload, sourceType: String, token: String?): RestOauth2AccessToken = requestAccessToken(payload: RequestTokenPayload, sourceType: String)

```

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OAuth2RestSourceAuthorizationService.kt
```diff
@@ -48,7 +48,7 @@ open class OAuth2RestSourceAuthorizationService(
 ) : RestSourceAuthorizationService {
     protected val httpClient = RestSourceAuthorizationService.httpClient()
 
-    override suspend fun requestAccessToken(payload: RequestTokenPayload, sourceType: String): RestOauth2AccessToken = withContext(Dispatchers.IO) {
+    override suspend fun requestAccessToken(payload: RequestTokenPayload, sourceType: String, token: String?): RestOauth2AccessToken = withContext(Dispatchers.IO) {
```
yatharthranjan: same as above

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/OuraAuthorizationService.kt
```diff
@@ -19,17 +19,23 @@ import org.radarbase.authorizer.api.RestOauth2AccessToken
...
+import org.slf4j.LoggerFactory
 import java.io.IOException
 
 class OuraAuthorizationService(
     @Context private val clients: RestSourceClientService,
     @Context private val config: AuthorizerConfig,
 ) : OAuth2RestSourceAuthorizationService(clients, config) {
-    override suspend fun requestAccessToken(payload: RequestTokenPayload, sourceType: String): RestOauth2AccessToken {
-        val accessToken: RestOauth2AccessToken = super.requestAccessToken(payload, sourceType)
+    override suspend fun requestAccessToken(payload: RequestTokenPayload, sourceType: String, token: String?): RestOauth2AccessToken {
```
yatharthranjan: same as above?

authorizer-app-backend/src/main/java/org/radarbase/authorizer/util/PkceUtil.kt
```diff
@@ -0,0 +1,20 @@
+package org.radarbase.authorizer.util
+
+import java.security.MessageDigest
+
+object PkceUtil {
+    // code verifier length needs to be between 43-128 characters
+    private const val CODE_VERIFIER_LENGTH = 64
```
yatharthranjan, line 7: can it be anything between 43-128 characters? or is 64 fixed for garmin?

authorizer-app-backend/src/main/resources/db/changelog/changes/20260204120000_add_code_verifier_column.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                        https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.10.xsd">
+
+    <changeSet author="aditya" id="20260204120000-add-code-verifier-column">
+        <addColumn tableName="registration">
+            <column name="code_verifier" type="varchar(128)">
```
yatharthranjan, line 10: presume this is only needed until auth code is obtained?

authorizer-app-backend/authorizer.yml
```diff
@@ -26,3 +26,16 @@ restSourceClients:
     clientId: <CLIENT_ID>
     clientSecret: <CLIENT_SECRET>
     scope: activity heartrate sleep profile
+    usesPkce: false
```
yatharthranjan: should we enable this automatically in the application when Garmin is set to oauth2.0? 
removes the chance of misconfiguration.

yatharthranjan (commented):
Thanks guys, nice work!. It looks already quite good, just some comments to improve robustness.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/enhancer/AuthorizerResourceEnhancer.kt
this-Aditya: Yeah, good point. Updated now.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/GarminOAuth2AuthorizationService.kt
this-Aditya: Updated, It does check this, now.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/GarminOAuth2AuthorizationService.kt
this-Aditya: I think we don't need to do this as it  is called by [reportDeregistration](https://github.com/RADAR-base/RADAR-Rest-Source-Auth/blob/94ada978fe4cdf883b37bc21d554dc416053d400/authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/SourceClientResource.kt#L148) which is triggered by garmin push service when the user has already revoked permissions on Garmin's side. 
We just need to clean up our own DB, not send the deregistration again.

authorizer-app-backend/src/main/resources/db/changelog/changes/20260204120000_add_code_verifier_column.xml
this-Aditya: Yeah, it is deleted after the flow is completed [here](https://github.com/RADAR-base/RADAR-Rest-Source-Auth/blob/2d73820dbceac7363b3d91986c23d01ca142cc0f/authorizer-app-backend/src/main/java/org/radarbase/authorizer/resources/RegistrationResource.kt#L170).

authorizer-app-backend/src/main/java/org/radarbase/authorizer/util/PkceUtil.kt
this-Aditya: No, it is not a garmin specific  requirement, just a choice.

authorizer-app-backend/authorizer.yml
this-Aditya: Okay, removed it from the yml config, but I think it's better to keep the field [here](https://github.com/RADAR-base/RADAR-Rest-Source-Auth/blob/2d73820dbceac7363b3d91986c23d01ca142cc0f/authorizer-app-backend/src/main/java/org/radarbase/authorizer/config/RestSourceClient.kt#L17), otherwise it would need to be recalculated [here](https://github.com/RADAR-base/RADAR-Rest-Source-Auth/blob/2d73820dbceac7363b3d91986c23d01ca142cc0f/authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/RegistrationService.kt#L24) every time. Keeping it in the config is fine since it's now set at runtime, regardless of what's in the yml.

PR #326, opened 2026-03-27
User correct user id endpoint for garmin oauth2

this-Aditya:
I noticed that Garmin uses different user ID endpoints for OAuth1 vs. OAuth2 flows. Right now, this is controlled only by code and not by external configuration.

Therefore, this PR updates the Garmin user ID to refer to the correct URL.


mpgxvii (approved):
LGTM! Did the other API endpoints stay the same?

this-Aditya:
Yes, the OAuth1 endpoint for retrieving the user ID is different, but it is handled by the [GarminOAuth1AuthorizationService](https://github.com/RADAR-base/RADAR-Rest-Source-Auth/blob/0bfd130284ff7e22835fc43b7ee9e8653aad727a/authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/GarminOauth1AuthorizationService.kt#L88). The authorization, token, and deregistration endpoints are also different and can be configured via the config, but this one cannot.

PR #328, opened 2026-03-27
fix: CI workflow docker image build

### LINKED ISSUE #327: CI workflow silently fails to build and push Docker images on dev/master
The CI workflow (`main.yml`) has been silently failing to build and push Docker images on `dev`/`master` pushes.

The  matrix JSON causes `fromJson()` to fail and the `docker` job gets skipped, but the workflow still shows as green because the other jobs (kotlin, node) pass fine.                                                                                                                             

See [this action run](https://github.com/RADAR-base/RADAR-Rest-Source-Auth/actions/runs/23605140210), all jobs passed but the docker job never ran and no images were pushed.                                                                                                                                                       
   
Also the image names in CI don't match the release workflow, so even when fixed, CI would push to a different package:
                                                                                                                                                                     
  | | `main.yml` | `release.yml` |
  |---|---|---|                                                                                                                                                      
  | Backend | `authorizer-app-backend` | `radar-rest-source-auth-backend` |
  | Frontend | `authorizer-app` | `radar-rest-source-authorizer` |

this-Aditya:
The `fromJson()` call in the `docker` job's matrix strategy was failing due to invalid JSON, causing the job to be skipped. The workflow still showed as green because the other jobs (kotlin, node) passed fine.                                                                                                                 
                                                                                                                                                                     
  Image name changes in `main.yml` for consistency:
  - `authorizer-app-backend` → `radar-rest-source-auth-backend`                                                                                                      
  - `authorizer-app` → `radar-rest-source-authorizer`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,13 +12,13 @@ env:
   REPOSITORY: ${{ github.repository }}
   IMAGES: >-
     [{
-    'name': 'authorizer-app-backend',
+    'name': 'radar-rest-source-auth-backend',
     'build_file': 'authorizer-app-backend/Dockerfile',
     'authors': 'Pauline Conde <pauline.conde@kcl.ac.uk>, Pim van Nierop <pim@thehyve.nl>',
     'description': 'RADAR-base rest sources authorizer backend application',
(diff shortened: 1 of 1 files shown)


mpgxvii (approved):
LGTM

PR #329, opened 2026-03-27
Sync dev with release to ensure the correct user ID endpoint is used in images pushed to dev.

this-Aditya:


gradle/libs.versions.toml
```diff
@@ -1,5 +1,5 @@
 [versions]
-project = "4.4.14"
+project = "4.4.15"
```
mpgxvii: Can we update to `4.4.16-SNAPSHOT`?

authorizer-app-backend/src/main/java/org/radarbase/authorizer/config/RestSourceClient.kt
```diff
@@ -15,7 +15,7 @@ data class RestSourceClient(
     val scope: String? = null,
     val state: String? = null,
     val usesPkce: Boolean = false,
-    val oauthVersion: String = "oauth2",
+    val oauthVersion: String = "oauth1",
```
mpgxvii: We'll use `oauth1` by default?

mpgxvii (dismissed):
LGTM! Just minor comments/questions.

gradle/libs.versions.toml
this-Aditya: Yeah.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/config/RestSourceClient.kt
this-Aditya: Let me change it to `oauth2` I just applied the [copilot review](https://github.com/RADAR-base/RADAR-Rest-Source-Auth/pull/324#discussion_r3001651314) for this PR.  😁

mpgxvii (approved):

PR #332, opened 2026-05-01
Add Google Health API authorization support

### LINKED ISSUE #331: Migrate source authorization from Fitbit OAuth to Google Health API OAuth
Fitbit's Web API is being sunset and replaced by the [Google Health API](https://developers.google.com/health/migration). Add support for the Google OAuth flow so users can authorize via Google OAuth instead of Fitbit's.

**Migration strategy:** When a user is linked to the Google Health source, if the same user already exists in the Fitbit source, a cascading delete is triggered to remove their Fitbit authorization upon successful Google OAuth completion.

this-Aditya:
A new authorization service handles Google's oauth2 flow.
Because the same participant may already have a Fitbit account from before the Google Health migration, the identity payload also carries an optional legacy Fitbit user ID, so we can keep both linked to the same RestSourceUser without losing history.                                           

PKCE is enabled for Google as an extra security layer on top of the standard oauth2 flow. If it turns out Google's auth servers don't play nicely with the PKCE flow in , I'll drop it and fall back to the plain oauth2 path.

### DIFF
--- a/authorizer-app-backend/authorizer.yml
+++ b/authorizer-app-backend/authorizer.yml
@@ -36,8 +36,8 @@ restSourceClients:
     clientId: <GARMIN_CLIENT_ID>
     clientSecret: <GARMIN_CLIENT_SECRET>
     oauthVersion: oauth2
-  # Google Health API - successor to the Fitbit Web API
-  - sourceType: GoogleHealth
+  # Google Health API
+  - sourceType: Google
     authorizationEndpoint: https://accounts.google.com/o/oauth2/v2/auth
     tokenEndpoint: https://oauth2.googleapis.com/token
     clientId: <GOOGLE_CLIENT_ID>
--- a/authorizer-app-backend/authorizer.yml
+++ b/authorizer-app-backend/authorizer.yml
@@ -26,11 +26,9 @@ restSourceClients:
     clientId: <CLIENT_ID>
     clientSecret: <CLIENT_SECRET>
     scope: activity heartrate sleep profile
-    usesPkce: false
   # Garmin OAuth2 PKCE configuration.
   # oauthVersion: "oauth2" selects the GarminOAuth2AuthorizationService;
   #               "oauth1" selects the legacy GarminOAuth1AuthorizationService.
-  # usesPkce: must be true for Garmin OAuth2 (Garmin requires PKCE).
   - sourceType: Garmin
     authorizationEndpoint: https://connect.garmin.com/oauth2Confirm
     tokenEndpoint: https://diauth.garmin.com/di-oauth2-service/oauth/token
@@ -38,4 +36,11 @@ restSourceClients:
     clientId: <GARMIN_CLIENT_ID>
     clientSecret: <GARMIN_CLIENT_SECRET>
     oauthVersion: oauth2
-    usesPkce: true
+  # Google Health API - successor to the Fitbit Web API
+  - sourceType: GoogleHealth
+    authorizationEndpoint: https://accounts.google.com/o/oauth2/v2/auth
+    tokenEndpoint: https://oauth2.googleapis.com/token
+    clientId: <GOOGLE_CLIENT_ID>
+    clientSecret: <GOOGLE_CLIENT_SECRET>
+    scope: https://www.googleapis.com/auth/googlehealth.activity_and_fitness.readonly https://www.googleapis.com/auth/googlehealth.health_metrics_and_measurements.readonly https://www.googleapis.com/auth/googlehealth.sleep.readonly https://www.googleapis.com/auth/googlehealth.settings.readonly https://www.googleapis.com/auth/googlehealth.profile.readonly
+    oauthVersion: oauth2
--- a/authorizer-app-backend/authorizer.yml
+++ b/authorizer-app-backend/authorizer.yml
@@ -42,5 +42,5 @@ restSourceClients:
     tokenEndpoint: https://oauth2.googleapis.com/token
     clientId: <GOOGLE_CLIENT_ID>
     clientSecret: <GOOGLE_CLIENT_SECRET>
-    scope: https://www.googleapis.com/auth/googlehealth.activity_and_fitness.readonly https://www.googleapis.com/auth/googlehealth.health_metrics_and_measurements.readonly https://www.googleapis.com/auth/googlehealth.sleep.readonly https://www.googleapis.com/auth/googlehealth.profile.readonly
+    scope: https://www.googleapis.com/auth/googlehealth.activity_and_fitness.readonly https://www.googleapis.com/auth/googlehealth.health_metrics_and_measurements.readonly https://www.googleapis.com/auth/googlehealth.sleep.readonly https://www.googleapis.com/auth/googlehealth.profile.readonly https://www.googleapis.com/auth/googlehealth.nutrition.readonly https://www.googleapis.com/auth/googlehealth.irn.readonly https://www.googleapis.com/auth/googlehealth.ecg.readonly
     oauthVersion: oauth2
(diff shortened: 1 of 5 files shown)


authorizer-app-backend/authorizer.yml
```diff
@@ -26,16 +26,21 @@ restSourceClients:
...
     oauthVersion: oauth2
-    usesPkce: true
+  # Google Health API
+  - sourceType: Google
+    authorizationEndpoint: https://accounts.google.com/o/oauth2/v2/auth
+    tokenEndpoint: https://oauth2.googleapis.com/token
+    clientId: <GOOGLE_CLIENT_ID>
+    clientSecret: <GOOGLE_CLIENT_SECRET>
+    scope: https://www.googleapis.com/auth/googlehealth.activity_and_fitness.readonly https://www.googleapis.com/auth/googlehealth.health_metrics_and_measurements.readonly https://www.googleapis.com/auth/googlehealth.sleep.readonly https://www.googleapis.com/auth/googlehealth.profile.readonly
+    oauthVersion: oauth2
```
yatharthranjan, line 46: is this required config? should it not be oauth2 by default?

authorizer-app-backend/authorizer.yml
```diff
@@ -26,16 +26,21 @@ restSourceClients:
...
     clientSecret: <GARMIN_CLIENT_SECRET>
     oauthVersion: oauth2
-    usesPkce: true
+  # Google Health API
+  - sourceType: Google
+    authorizationEndpoint: https://accounts.google.com/o/oauth2/v2/auth
+    tokenEndpoint: https://oauth2.googleapis.com/token
+    clientId: <GOOGLE_CLIENT_ID>
+    clientSecret: <GOOGLE_CLIENT_SECRET>
+    scope: https://www.googleapis.com/auth/googlehealth.activity_and_fitness.readonly https://www.googleapis.com/auth/googlehealth.health_metrics_and_measurements.readonly https://www.googleapis.com/auth/googlehealth.sleep.readonly https://www.googleapis.com/auth/googlehealth.profile.readonly
```
yatharthranjan: what about other scopes like - nutrition, irn, ecg, location (some studies have been explicitly requesting location records for activities recently so would be good to include if possible)?

yatharthranjan (commented):
Thanks @this-Aditya, looks nice. Has this been tested working? If so, can we release this soon (so at least we can authorise new participants on this, we can always pull the data later)

authorizer-app-backend/authorizer.yml
this-Aditya: Yeah, it is [OAuth 2.0 by default](https://github.com/RADAR-base/RADAR-Rest-Source-Auth/blob/012d44a96cc376c0bb5987b72da43183f18d457c/authorizer-app-backend/src/main/java/org/radarbase/authorizer/config/RestSourceClient.kt#L18). I only kept it here to highlight it.

authorizer-app-backend/authorizer.yml
this-Aditya: Yeah, we can add these, but google docs says that we can only request the scopes for which we are collecting data. So, if we want to increase the number of data points, we can add them

authorizer-app-backend/src/main/java/org/radarbase/authorizer/enhancer/AuthorizerResourceEnhancer.kt
```diff
@@ -112,8 +114,9 @@ class AuthorizerResourceEnhancer(
         bind(DelegatedRestSourceAuthorizationService::class.java)
             .to(RestSourceAuthorizationService::class.java)
 
-        // Bind Garmin service based on a configured oauthVersion: "oauth2" → PKCE flow, "oauth1" → legacy flow.
-        if (sourceTypeOauthMap[GARMIN_AUTH].equals("oauth2", ignoreCase = true)) {
+        // Bind Garmin service based on configured oauthVersion: "oauth2" → PKCE flow, "oauth1" → legacy flow.
+        val garminUsesPkce = restSourceClients.clients.firstOrNull { it.sourceType == GARMIN_AUTH }?.usesPkce == true
+        if (garminUsesPkce) {
```
yatharthranjan, line 119: I think more readable to just use oauthVersion from config here

authorizer-app-backend/src/main/java/org/radarbase/authorizer/enhancer/AuthorizerResourceEnhancer.kt
yatharthranjan, line 133: should we somehow mark this as deprecated?

this-Aditya: It would be nice if we could do this after september 2026, as they will still receive data from the fitbit app until september if they signed up using their google account.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/enhancer/AuthorizerResourceEnhancer.kt
this-Aditya: I've updated this in PR #333. As this PR derives `usesPkce` from the `RestSourceClient`, I think that's a better place to handle it than doing it here.

authorizer-app-backend/authorizer.yml
yatharthranjan: likely remove it if that is the default and does not require configuration

authorizer-app-backend/authorizer.yml
yatharthranjan: yes shoudl at least add nutrition, irn and ecg

authorizer-app-backend/authorizer.yml
this-Aditya: Should I do it now, or after the Garmin migration? We might be able to remove this configuration completely then, since we'll only be left with oauth 2.0.

authorizer-app-backend/authorizer.yml
this-Aditya: okay

authorizer-app-backend/src/main/java/org/radarbase/authorizer/enhancer/AuthorizerResourceEnhancer.kt
yatharthranjan: I think ok to mark as deprecated (not decomissioned), with details on when it will be non-functional (sept 2026)

authorizer-app-backend/authorizer.yml
yatharthranjan: yes i think for garmin, it should still be configurable if needed (especially for legacy deployments and until Garmin forces use of OAuth2.0)

yatharthranjan (dismissed):
LGTM, please merge the base branch (outdated atm)

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/DelegatedRestSourceAuthorizationService.kt
```diff
@@ -63,5 +63,6 @@ class DelegatedRestSourceAuthorizationService(
         const val GARMIN_AUTH = "Garmin"
         const val FITBIT_AUTH = "FitBit"
         const val OURA_AUTH = "Oura"
+        const val GOOGLE_AUTH = "Google"
```
mpgxvii: Do you think we should keep this `Google` or `GoogleHealth`? The implementation is mostly health but the auth is generic Google.

this-Aditya: Let me update it. I was thinking of having a single authorization flow if we expand Google support for other use cases, but since we might need different scopes, a separate flow should be fine.

mpgxvii (approved):
Thanks! LGTM.

PR #333, opened 2026-05-29
Make usesPkce a derived property on RestSourceClient

this-Aditya:
Derive `usesPkce` in rest source client and remove it from constructor param so it can never be added to the configs.

### DIFF
--- a/authorizer-app-backend/src/main/java/org/radarbase/authorizer/config/RestSourceClient.kt
+++ b/authorizer-app-backend/src/main/java/org/radarbase/authorizer/config/RestSourceClient.kt
@@ -16,20 +16,11 @@ data class RestSourceClient(
     val grantType: String? = null,
     val scope: String? = null,
     val state: String? = null,
-    val oauthVersion: String = "oauth2",
+    val oauthVersion: OAuthVersion = OAuthVersion.OAUTH2,
 ) {
-    val usesOauth2: Boolean
-        get() = when {
-            oauthVersion.equals("oauth2", ignoreCase = true) -> true
-            oauthVersion.equals("oauth1", ignoreCase = true) -> false
-            else -> throw IllegalArgumentException(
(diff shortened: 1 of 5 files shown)


authorizer-app-backend/src/main/java/org/radarbase/authorizer/config/RestSourceClient.kt
```diff
@@ -14,9 +16,24 @@ data class RestSourceClient(
     val grantType: String? = null,
     val scope: String? = null,
     val state: String? = null,
-    val usesPkce: Boolean = false,
     val oauthVersion: String = "oauth2",
```
yatharthranjan: if it only supports 2 values, would it be better as an enum class then ?

this-Aditya: Yeah, good point. 
 Would it be nice, if we just keep this as boolean instead, something like `isOauth2` to keep it simple?

mpgxvii: Actually maybe enum might be cleaner?
```
OAuthVersion.OAUTH1
OAuthVersion.OAUTH2
```

mpgxvii (dismissed):
LGTM, just a minor commonent on the oauth version. Thanks!

authorizer-app-backend/src/main/java/org/radarbase/authorizer/config/RestSourceClient.kt
this-Aditya: Updated!!

mpgxvii: Thank you!

mpgxvii (approved):

PR #334, opened 2026-06-01
Make add_code_verifier changeset idempotent to fix staging re-run

this-Aditya:
While testing on stage, I noticed the last deployed changeset id contained `.xml`, while the current changelog id does not. Liquibase treat it as a new changeset and re-run the `addColumn` on the next deployment, failing  with `column "code_verifier" already exists` and blocking app startup.

This PR adds a `preConditions` check: if the `code_verifier` column already  exists, mark the changeset as run  instead of failing. This `.xml`-suffixed id was only deployed to staging it was never deployed to production. So this fix specifically unblocks the staging deployment.


yatharthranjan (approved):
LGTM, do you know how the ID changed?

mpgxvii (approved):

PR #336, opened 2026-06-08
Rename source type to GoogleHealth in configs

this-Aditya:
As per the discussion of PR: https://github.com/RADAR-base/RADAR-Rest-Source-Auth/pull/332, it is decided to rename the source type from `Google` to `GoogleHealth`. But I missed to update this in configs. This PR just does the remaining update.


mpgxvii (approved):
LGTM, thanks!

PR #339, opened 2026-06-29
Manage Google Health subscriptions in Rest-Source-Auth

this-Aditya:
Moves per-user Google Health webhook subscriptions out of the Push Endpoint and into the authorizer, which already owns each user's authorization state. A subscription is now created or removed synchronously as a user authorizes, refreshes, revokes, or is deleted, with the state kept in a new one-to-one `rest_source_user_subscription` table. It also adds a manual subscribe/unsubscribe endpoint.


yatharthranjan (approved):
LGTM

PR #344, opened 2026-07-20
feat: dexcom initial integration

Chin-eng:


this-Aditya:
Hi @Chin-eng, I see that the CI is failing. Could you please run ktlint format and push the changes again?

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/DexcomAuthorizationService.kt
```diff
@@ -0,0 +1,133 @@
...
+class DexcomAuthorizationService(
+    @Context private val clients: RestSourceClientService,
+    @Context private val config: AuthorizerConfig,
+) : OAuth2RestSourceAuthorizationService(clients, config) {
+
+    override suspend fun requestAccessToken(
+        payload: RequestTokenPayload,
+        sourceType: String,
+        token: String?,
+    ): RestOauth2AccessToken = withContext(Dispatchers.IO) {
```
this-Aditya, line 41: This version of `requestAccessToken` is meant to be overridden when the flow needs the state token. Garmin and GoogleHealth use it to look up the pkce code verifier. This service doesn't use pkce and doesn't use the `token` parameter, so I'd recommend overriding the two-parameter version instead.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/DexcomAuthorizationService.kt
```diff
@@ -0,0 +1,133 @@
...
+import org.radarbase.jersey.exception.HttpBadGatewayException
+import org.slf4j.LoggerFactory
+
+/**
+ * Dexcom OAuth2 authorization service.
+ *
+ * Dexcom requires client_id and client_secret as form parameters in the token
+ * request rather than HTTP Basic authentication.
+ */
+class DexcomAuthorizationService(
```
this-Aditya, line 33: I see that `revokeToken` isn't overridden here. As you mention in the class doc, Dexcom doesn't work with basic credentials, but the parent class method uses exactly that. Even though Dexcom has no revocation endpoint, this method still needs to be overridden so that it doesn't perform the wrong action and. You could either return false or throw exception.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/DexcomAuthorizationService.kt
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    companion object {
+        private val logger = LoggerFactory.getLogger(DexcomAuthorizationService::class.java)
+    }
+
+    @Serializable
+    private data class DexcomDataRangeResponse(
+        @SerialName("userId") val userId: String,
+    )
```
this-Aditya: If we're only extracting the user ID, the name `DexcomDataRangeResponse` doesn't align with what's actually being extracted. I'd recommend skipping this class and using [`RestOauth2UserId`](https://github.com/RADAR-base/RADAR-Rest-Source-Auth/blob/c4187d0360d937ec2a5eeb7c83e0eb937cda6f5a/authorizer-app-backend/src/main/java/org/radarbase/authorizer/api/ApiDeclarations.kt#L54) directly, it already does the same thing.

docker/etc/rest-source-authorizer/authorizer.yml
```diff
@@ -23,6 +23,12 @@ redis:
   uri: redis://redis:6379
 
 restSourceClients:
+  - sourceType: Dexcom
+    authorizationEndpoint: https://sandbox-api.dexcom.com/v3/oauth2/login
+    tokenEndpoint: https://sandbox-api.dexcom.com/v3/oauth2/token
```
this-Aditya, line 28: Please update the sandbox URLs if the real ones are ready.

this-Aditya (commented):
Hi Chinee, could you please take a look at the inline comments?

docker/etc/rest-source-authorizer/authorizer.yml
Chin-eng: The production access has not granted yet. I will update it once the dexcom approves the request.

authorizer-app-backend/src/main/java/org/radarbase/authorizer/service/DexcomAuthorizationService.kt
```diff
@@ -100,6 +101,11 @@ class DexcomAuthorizationService(
         }
     }
 
+    override suspend fun revokeToken(user: RestSourceUser): Boolean {
+        logger.info("Token revocation not supported for Dexcom")
```
this-Aditya, line 105: `logger.error` would be better here, or we could also throw an exception.

this-Aditya (commented):
LGTM, Thanks

this-Aditya (approved):

